PR #26, opened 2023-10-14
Feat/add s3 backend

### LINKED ISSUE #10: add native s3 storage backend
I was having a look at implementing it myself, but found the backend traits to not be async.  I also found in the changelog, that async was removed in 0.4.  Is this going to change soonish again?  If so, I'll defer it.

I would use `rust-s3`, because afaict the official aws sdk is async only.  What do you think?

Scandiravian:
This PR adds an S3 backend which improves the drop-in-ability of rustic as a replacement for Restic.

When reviewing the code I highly recommend doing it one commit at a time, since some of these will be dropped before the PR is marked as ready. I have included them to show how I have tested the code.

When the PR is ready for review I will also squash the undropped commits into a single one, as per the contribution guidelines.

I haven't tested the changes since the original PR to the rustic repo (https://github.com/rustic-rs/rustic/pull/871) , but unless there has been some major changes to how backends work, the testing I did back then should still be valid. I have some thoughts on how integration tests could be added, which I'll probably share in an issue instead to not bloat the scope of this PR too much.

closes #10


src/backend/s3.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl ReadBackend for S3Backend {
+    fn location(&self) -> String {
+        let mut location = "s3:".to_string();
+        location.push_str(&self.path_style_url());
+        location
+    }
+
+    fn set_option(&mut self, _option: &str, _value: &str) -> RusticResult<()> {
+        todo!()
```
Scandiravian, line 223: I'm not sure what should be implemented here, so some help on this would be really appreciated

aawsome: Actually, you can keep this empty, i.e. maybe just print a warning or return a error for each option/value that is given. If you have extra options you want to set via the config, you can evaluate them here...

aawsome:
@Scandiravian Please note that in #73 I propose to completely change the 
Repository API.

Also, in that PR I was able to suprisingly easy add an `opendal` backend - which supports S3 and many other storages. To use opendal, you specify the service as as repository, like `repository = "opendal:s3"` and specify all other information (if needed by the opendal service) in the `repository.options` setting, see e.g. https://opendal.apache.org/docs/services/s3#configuration for S3 config options.

In case the opendal S3 backend implementation supports all features we want to have, I would favor that solution to a specific s3 backend implementation.
Feel free to try #73 - it compiles with the latest `rustic` code, you just need to specify the correct ependency for rustic_core.

aawsome:
Just wanted to mention that #73 is nearly finished from my side - and already adds the possibility to use something like "s3:https://s3.amazonaws.com/bucket" as repository.

Scandiravian:
@aawsome great work! Should I close this PR in favour of #73?

aawsome:
@Scandiravian Please try out the S3 backend provided by #73. If this suits your needs, I think we should favor that and close this PR.

But anyway, help with tests is highly appreciated and should be now possible using in-memory backends or other mocks...

Issue #27, opened 2023-10-23
rustic_core-0.1.1 fails building on OpenBSD


bket:
Building rustic-0.6.0 fails on OpenBSD current, which is related to rustic_core. Building rustic_core-0.1.1 on OpenBSD current results in the error below.

This issue is related to https://github.com/rustic-rs/rustic/issues/917.

```
   Compiling rustic_core v0.1.1 (/home/code/rustic_core)
error[E0599]: no variant or associated item named `FromErrnoError` found for enum `LocalErrorKind` in the current scope
   --> src/backend/local.rs:666:38
    |
666 |             .map_err(LocalErrorKind::FromErrnoError)?;
    |                                      ^^^^^^^^^^^^^^
    |                                      |
    |                                      variant or associated item not found in `LocalErrorKind`
    |                                      help: there is a variant with a similar name: `FromTryIntError`
    |
   ::: src/error.rs:598:1
    |
598 | pub enum LocalErrorKind {
    | ----------------------- variant or associated item `FromErrnoError` not found for this enum

error[E0599]: no variant or associated item named `FromErrnoError` found for enum `LocalErrorKind` in the current scope
   --> src/backend/local.rs:706:38
    |
706 |             .map_err(LocalErrorKind::FromErrnoError)?;
    |                                      ^^^^^^^^^^^^^^
    |                                      |
    |                                      variant or associated item not found in `LocalErrorKind`
    |                                      help: there is a variant with a similar name: `FromTryIntError`
    |
   ::: src/error.rs:598:1
    |
598 | pub enum LocalErrorKind {
    | ----------------------- variant or associated item `FromErrnoError` not found for this enum

error[E0599]: no variant or associated item named `SettingFilePermissionsFailed` found for enum `LocalErrorKind` in the current scope
   --> src/backend/local.rs:749:42
    |
749 |                 .map_err(LocalErrorKind::SettingFilePermissionsFailed)?;
    |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ variant or associated item not found in `LocalErrorKind`
    |
   ::: src/error.rs:598:1
    |
598 | pub enum LocalErrorKind {
    | ----------------------- variant or associated item `SettingFilePermissionsFailed` not found for this enum

error[E0599]: no variant named `SymlinkingFailed` found for enum `LocalErrorKind`
   --> src/backend/local.rs:918:78
    |
[...]

PR #28, opened 2023-10-24
Unbreak building of rustic_core-0.1.1 on OpenBSD (#27)

bket:
Issue was caused by a couple of configuration checks.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -14,7 +14,7 @@ use std::{
     time::SystemTimeError,
 };
 
-#[cfg(not(any(windows, target_os = "openbsd")))]
+#[cfg(not(windows))]
 use nix::errno::Errno;
 
 use aes256ctr_poly1305aes::aead;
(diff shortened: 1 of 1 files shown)


bket:
Forgot to mention that with this patch all tests provided by `cargo test` pass.

bket:
@simonsan, @aawsome, clippy seems to fail at a point that has not been touched by this PR. Is this a known issue?

aawsome:
I would say it is definitively a new Rust/Clippy version which gives these warnings - I'll work on a fix and then we can merge this PR.
Sorry for being a bit short of time ATM - day job and family are currently a bit demanding...

aawsome:
I would say it is definitively a new Rust/Clippy version which gives these warnings - I'll work on a fix and then we can merge this PR.
Sorry for being a bit short of time ATM - day job and family are currently a bit demanding...

aawsome (approved):
LGTM!
Thanks a lot @bket and sorry for the inconveniences!

PR #74, opened 2023-11-29
feat: add `--ignore-file` command line flag

ProspectPyxis:
This adds the `--ignore-file` command line option to treat each listed file like a `.gitignore`. This is distinct from the `--exclude-file` option, since it uses the walk builder and thus will match the filename even in subdirectories.

The name of the flag could be changed - as it stands, I don't think it's very clear from a glance what the difference between this flag and `--exclude-file` is - but this name seems like a good enough choice for now.

Closes rustic-rs/rustic#832.

### DIFF
--- a/src/backend/ignore.rs
+++ b/src/backend/ignore.rs
@@ -100,6 +100,11 @@ pub struct LocalSourceFilterOptions {
     #[cfg_attr(feature = "merge", merge(strategy = merge::bool::overwrite_false))]
     pub no_require_git: bool,
 
+    /// Treat the provided filename like a .gitignore file (can be specified multiple times)
+    #[cfg_attr(feature = "clap", clap(long, value_name = "FILE"))]
+    #[cfg_attr(feature = "merge", merge(strategy = merge::vec::overwrite_empty))]
+    pub ignore_file: Vec<String>,
+
     /// Exclude contents of directories containing this filename (can be specified multiple times)
     #[cfg_attr(feature = "clap", clap(long, value_name = "FILE"))]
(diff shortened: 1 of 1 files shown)


aawsome:
Thanks @ProspectPyxis for the PR!
I think this solves https://github.com/rustic-rs/rustic/issues/832 directly once we publish a new rustic_core version and bumb the rustic dependency.

I have one thing: Is the name of the option a good choice? I personally think that `--ignore-filename` could be also a better fit, but I am open to discuss it...

ProspectPyxis:
Yes, I agree - the name feels a little awkward as it stands. There's not exactly a good short name that encapsulates "treat this file as a gitignore, so this works even in subfolders"...

I can make a quick commit to fix this if you believe that would be a better name.

aawsome:
@simonsan do you have a better idea?

ProspectPyxis:
I like `--ignore-source`, personally. I don't think we can fully avoid ambiguity here, but that's pretty understandable to me as "use this file as an ignore source". I could also update the documentation to refer to a filename rather than a file, just to be even more clear.

aawsome:
I'd like to throw in `--use-ignorefile` or `--custom-ignorefile` to make clear that we are talking about ignorefiles and not files to directly ignore. 
I'll keep thinking about what would be a good and least misleading option name ;-)

aawsome:
> So for that you would add another `--<specify-ignore-file-command> <profile-dependent-ignore-file-path>`.

Yes, this is how clap works in that case.

As a didn't hear arguments against `--custom-ignorefile`, let's just use this and complete this PR!

ProspectPyxis:
Updated as requested!

PR #94, opened 2023-12-15
better copy progress

aawsome:
is based on #73 (to minimize merge conflicts) and should be merged after it.

closes https://github.com/rustic-rs/rustic/issues/960


aawsome (commented):
LGTM

PR #97, opened 2023-12-16
backup: run size scanning parallel to backup; add no-scan option

aawsome:
Breaking change: This removes the `impl Iterator` from `LocalSource`. To still be able to iterate over its entries, the `ReadSource` trait (and its `entries()` method) is now public.

The neccessary change for rustic is in https://github.com/rustic-rs/rustic/pull/966

closes  https://github.com/rustic-rs/rustic/issues/689

This PR is based on #73 which should be merged first


aawsome:
https://github.com/rustic-rs/rustic_core/pull/97 adds the no_scan option and by default starts the backup immediately (running the scan in parallel) if no `--no-scan` option is given.
To have this in rustic, we need to first bump the rustic_core version to a version which includes the mentioned PR.

aawsome (commented):
LGTM

PR #100, opened 2023-12-20
cache: Don't write warnings if cache files don't exist

aawsome:
see https://github.com/rustic-rs/rustic/issues/970


aawsome (commented):
LGTM

PR #110, opened 2023-12-30
feat: Add tls-native, tls-rustls features to allow building w/o Rustls

jirutka:


aawsome:
Agree to @simonsan - when the backend refactor is done, the TLS things will be all in the crate rustic_backends, so we better wait for the refactoring #73 to land!

Besides this, if there are use cases where you need other TLS options, I would very much agree to adding the possibilites as features. It might be that the depending crate(s) (e.g. opendal)  need to support those features, too...

jirutka:
> What's your use case?

Packaging rustic for Alpine Linux.

> Why is this change needed and what does it improve/solve?

Bundling (static linking) TLS library is a very bad idea from a security perspective, and Linux distributions try to avoid it. Also, rustls/ring doesn’t support all platforms supported by Rust, e.g. ppc64le and s390x.

jirutka:
> if the features are mutually exclusive, could you throw a compile error in lib.rs, in case both are turned on?

Sure, done.

aawsome:
Now that #73 is merged, this PR can continue...
@jirutka: Your changes should now go to `crates/backend`. Can you move this or do you need support?

Also, now that we have added opendal, this should be also treated - however, I don't know if this is already supported by opendal; this may need a bit of research.

PR #126, opened 2024-01-18
backend: Add sftp backend

aawsome:
Note that sftp in opendal is currently only supported on unix, see https://github.com/apache/incubator-opendal/issues/2963


aawsome (commented):
LGTM

PR #138, opened 2024-01-25
fix: don't abort on negative elapsed time in packer/indexer

aawsome:
closes https://github.com/rustic-rs/rustic/issues/1018


aawsome (commented):
LGTM

PR #154, opened 2024-02-04
feat: Add extra check before writing data

### LINKED ISSUE #153: Implement verification of blobs before uploading
**Introduction:**
Recently, users have encountered issues with corrupted data when using the "max" compression feature in the (go) tool restic. These instances of data corruption can lead to data loss and other complications, highlighting the need for a robust identification of such errors beside the fix itself to detect similar bugs. The PR of restics solution is linked below.

**Explanation of the Feature:**
The proposed feature involves implementing this blob verification before uploading data. This process would detect errors and inconsistencies in the data, preventing corrupted data from being uploaded to the repository. By performing verification checks prior to uploading, rustic can ensure the integrity of stored data and provide users with greater confidence in their backups. Today a corruption is not detectable until the user runs the "read-data" command, which may not that often or sometimes never.

As mentioned in the discussion: I would be nice "to have this as optional feature which can switched on using CLI options or the config file." I personally would suggest a default activation would be great, because on a backup solution security and verification should be the default.

**Benefits:**
* Prevents data loss by detecting errors before they are uploaded to the repository.
* Provides early detection of data corruption, allowing users to take corrective actions promptly.
* Saves bandwidth by avoiding the upload of corrupted data.

**Disadvantages:**
Possible disadvantage mentioned in the rustic discussion (linked below):
> I think the implementation would be not too hard, but might slow down backup runs a bit...

**References:**
* [Discussion on rustic repo](https://github.com/rustic-rs/rustic/discussions/1049)
* [Restic Forum post about the problem](https://forum.restic.net/t/check-reports-errors-wrong-data-returned/7201/7)
* [Github PR of restic](https://github.com/restic/restic/pull/4681)

---
Hope this is written well enough, thank you for your great work and fast support! <3

aawsome:
This PR adds by default an extra check on compressed/encrypted data before it is written to the repository by decomressing/decrypting and comparing it with the original data.

The check can be opted-out by using `rustic init --set-extra-check=false` (or using the `config` command).

closes #153 

TODO:
- [x] Handle options for other commands which write blobs: `copy`, `merge`, `prune`, `repair snapshot` -> currently this is no option, but hardcoded
- [x] Decide about where to put the option: To individual command options? To general (rustic) options? Into the repostiory config?
- [x] Decide about default if not specified by the user
- [x] Add checks about pack headers - No, headers are not compressed!
- [x] Add checks about other repository files
- [x] Decide name and do we want to have it positive (like `--extra-check`) or negative (like `--no-extra-check`)


simonsan:
> I did some performance tests and must say that the extra verification step would need quite some extra CPU, about 50-80% increase and also implies an increase in the elapsed time. Based on this I would tend to making this optional, i.e. let the default be to not do the extra checks.

I agree, it should be optional.

> About where to put the options, I tend to putting it to the repository config.

Yeah, I mean, it's a question where people will use it. For example, if someone would want to use it only to copy or to back up, it would be probably better to not have it as a global option. Although I'm not sure, if it would be making sense to use such an option when copying?

aawsome:
> Yeah, I mean, it's a question where people will use it. For example, if someone would want to use it only to copy or to back up, it would be probably better to not have it as a global option. Although I'm not sure, if it would be making sense to use such an option when copying?

I think it is an option about quality checks done to data added to the repository. It it is set, people know that all data (by all commands) added to the repo did get that extra-check. If not, then they know they might want to do an extra `check -- read-data` if they fear for data which was already corrupt while adding. This is why I think an option in the repository config which can and will be usually set during `init` is the right place.

duracell:
> I did some performance tests and must say that the extra verification step would need quite some extra CPU, about 50-80% increase and also implies an increase in the elapsed time.

How did you measure this? I tried to do performance test on restic and there the new version (with their linked implementation) only has a ~5% increase in time (1 CPU tested, nearly 100% all the time). I tested: "restic 0.16.4 compiled with go1.21.6 on linux/amd64" (normal and with "--no-extra-verify") and "restic 0.16.3 compiled with go1.21.6 on linux/amd64".
Or is this normal, that for restic the impact is so much smaller?

> Based on this I would tend to making this optional, i.e. let the default be to not do the extra checks.

I understand this. I already said that from my opinion the integrity should be the main goal, but I totally understand if you see the risks and cost-benefit relation different then me. For myself it's totally fine as an option, I just think about the user which didn't look at all of the options and if they want the best data integrity instead of performance.

> About where to put the options, I tend to putting it to the repository config - this also automatically solves the question about the other commands.

Sounds good for me.

simonsan:
> I understand this. I already said that from my opinion the integrity should be the main goal, but I totally understand if you see the risks and cost-benefit relation different then me. For myself it's totally fine as an option, I just think about the user which didn't look at all of the options and if they want the best data integrity instead of performance.

Whenever we have a way to create a config interactively, that would be a question, that would be asked to the user and different default settings being written to the config. E.g. the tradeoff between data integrity, size, and performance of backups. It is true though, that I personally also value data integrity over performance (it's a backup in the end), so while it might make sense to not set it as a default. It might be worth it to actually update all the available configuration files with that setting, to let people know about it, set it there as a default and directly warn about performance impact.

aawsome:
> How did you measure this? I tried to do performance test on restic and there the new version (with their linked implementation) only has a ~5% increase in time (1 CPU tested, nearly 100% all the time). I tested: "restic 0.16.4 compiled with go1.21.6 on linux/amd64" (normal and with "--no-extra-verify") and "restic 0.16.3 compiled with go1.21.6 on linux/amd64".
> Or is this normal, that for restic the impact is so much smaller?

I actually tested with restic as I wasn't 100% sure about the WIP implementation in rustic.

I used an empty local repository on `/tmp/repo` as I only wanted to measure a command which main work is about adding blobs.
Then,
```
dd if=/dev/urandom bs=1M count=50 | /usr/bin/time restic backup -r /tmp/repo/ --stdin
```
with `--no-extra-verify` I got `1.83user 0.30system 0:01.26elapsed 168%CPU (0avgtext+0avgdata 141512maxresident)k`
without I got `3.26user 0.38system 0:01.41elapsed 257%CPU (0avgtext+0avgdata 196468maxresident)k `

aawsome:
@duracell Thanks for your results. Yes, indeed, this could be CPU-specific depending on how much hardware-accelerated the decryption and decompression is. 

I'm open to discuss what should be the default.
However, I think we must differentiate potential errors which are caught by this extra check. restic has this blob in-memory, then first hashes it, then compresses it and finally encrypts it. If the check is activated, the final result is also decrypted, decompressed and again hashed to check if the hashes match.
As decryption/encryption and hashing never poses any problem, this effectively only tests if compressing followed by decompressing still yields the present data.
As I already explained, restic introduced this step as it has good reasons for questioning that this is really valid. Rustic IMO doesn't have this reason.

Moreover, we specify *both*  the compression level *and* this extra checking possibility in the config file. So, if the high-on-edge compression levels should also pose this problems in rustic, people can add both in one command by using `rustic init --set-compression 22 --set-extra-check true`

aawsome:
> Edit: After further testing I saw that in all tests the compression was on max settings, while likely on auto on yours.
> I switched it to auto and then in the tests with a complete random file it was a max 30% increase in time. So with the auto setting it is a big(ger) time cost. I tested it (on only a single run) with the complete backup again and with auto it was 20% longer run with verify.

Ah, yes, of course. Decompressing is much cheaper than compressing - especially for high compression rates. So for high compression the extra check wouldn't add as much relative overhead.

duracell:
> As I already explained, restic introduced this step as it has good reasons for questioning that this is really valid. Rustic IMO doesn't have this reason.

But why doesn't rustic have this reason? Just because it never happened? Or is there any specific step which rustic has that restic is missing?
 
> Moreover, we specify _both_ the compression level _and_ this extra checking possibility in the config file. So, if the high-on-edge compression levels should also pose this problems in rustic, people can add both in one command by using `rustic init --set-compression 22 --set-extra-check true`

Yeah, I think this is fine. As mentioned it's a matter of priority and cost-benefit, I would do the same, just with the restic way of the disable-verify option, because then people who don't care to optimize/understand the config have better checks, but slower backups.
But for me it's totally valid to implement it like you specified. There is nothing wrong about this. I'm just a person who wants to have as much protection as possible for my backups (and also because restic had this problems with my configuration and that scared me a lot!).

aawsome:
> But why doesn't rustic have this reason? Just because it never happened?

First, it should be the job of an compression library to ensure that no data is corrupted and not rustic's job. I mean, there are so many things which could go wrong in theory (and for sure some will go wrong in practice) which are also not double-checked. For the simple reason that we must trust that some things just work as specified.

The reason for restic to add this compression/decompression check is that data corruption did happen (in oct 2023), got fixed and it happened a second time (in feb 2024). Both times the bug was found by restic users. This indicates that the tests which were run on the used compression library are obviously not suitable to find these kind of bugs - so it is very reasonable for restic to add extra checks.

Rustic, on the other side, uses a completely different compression library, https://github.com/facebook/zstd. I won't say that these kind of problems cannot occur with that library and in fact, they had this problem (in april 2023) , see https://github.com/facebook/zstd/releases/tag/v1.5.5.
However this seems to be fixed and didn't appear since then - though I am pretty sure about massive fuzzing tests done since then.

My conclusion is that I trust the compression library rustic uses more than the one restic uses. But of course everybody is free to draw their own conclusions, therefore I am definitively willing to finish this feature soon which allows people to choose to use this check!

aawsome:
> And why proposal for the check with higher compression. I know it was the max setting for restic, but it a bug could have been in every setting, couldn't it?

All known bugs of this kind (1 in the facebook library, 2 in the go library) occurred for higher compression levels.

aawsome:
Now that the check is also implemented for snpashot/index files, I ran some comparison using hyperfine:
```
alex-dev@latitude:~/rust/rustic$ hyperfine -p "rm -rf /tmp/repo" "dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init -"
Benchmark 1: dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init -
  Time (mean ± σ):      1.315 s ±  0.052 s    [User: 1.254 s, System: 0.448 s]
  Range (min … max):    1.242 s …  1.430 s    10 runs
 
alex-dev@latitude:~/rust/rustic$ hyperfine -p "rm -rf /tmp/repo" "dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-extra-check -"
Benchmark 1: dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-extra-check -
  Time (mean ± σ):      1.337 s ±  0.038 s    [User: 1.359 s, System: 0.482 s]
  Range (min … max):    1.281 s …  1.380 s    10 runs
```
and, using maximum compression, surprisingly:
```
alex-dev@latitude:~/rust/rustic$ hyperfine -p "rm -rf /tmp/repo" "dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-compression 22 -"
Benchmark 1: dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-compression 22 -
  Time (mean ± σ):     13.075 s ±  0.735 s    [User: 26.964 s, System: 22.980 s]
  Range (min … max):   12.003 s … 14.605 s    10 runs
 
alex-dev@latitude:~/rust/rustic$ hyperfine -p "rm -rf /tmp/repo" "dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-compression 22 --set-extra-check -"
Benchmark 1: dd if=/dev/urandom bs=1M count=50 | ./target/release/rustic backup --init --set-compression 22 --set-extra-check -
  Time (mean ± σ):     12.607 s ±  0.564 s    [User: 26.448 s, System: 21.983 s]
  Range (min … max):   11.771 s … 13.424 s    10 runs
```
(I never used max compression level myself, but now I know why it should be discouraged...)

TL;DR: On rustic, the overhead is much smaller than on restic. (I can't explain the max compression result, maybe this is due to the fact that parallelization is not at its optimum yet or due to other side effects.

I also realized that in contrast to restic, there is almost no memory overhead when activating the extra check.

aawsome:
After the benchmarks I've seen I decided to make the check the default and allow to opt-out.

crates/core/src/backend/decrypt.rs
```diff
@@ -356,8 +361,19 @@ impl<C: CryptoKey> DecryptBackend<C> {
...
             key,
             zstd: None,
+            extra_check: false,
         }
     }
+
+    fn decrypt_file(&self, data: &[u8]) -> RusticResult<Vec<u8>> {
+        let decrypted = self.decrypt(data)?;
+        Ok(match decrypted.first() {
+            Some(b'{' | b'[') => decrypted, // not compressed
```
simonsan, line 371: Would be determining compression by magic number (zstd is `0xFD2FB528` i think?) better here?

From this issue: https://github.com/facebook/zstd/issues/768

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
             }
             None => self.key().encrypt_data(data)?,
         };
-        let id = hash(&data);
-        self.write_bytes(tpe, &id, false, data.into())
+        if self.extra_check {
+            let check_data = self.decrypt_file(&data_encrypted)?;
+            if data != check_data {
+                return Err(CryptBackendErrorKind::ExtraCheckFailed.into());
```
simonsan, line 417: Would rather call it `integrity check` or something?

```suggestion
                return Err(CryptBackendErrorKind::IntegrityCheckFailed.into());
```

crates/core/src/backend/decrypt.rs
```diff
@@ -334,6 +337,8 @@ pub struct DecryptBackend<C: CryptoKey> {
     key: C,
     /// The compression level to use for zstd.
     zstd: Option<i32>,
+    /// The whether do do an extra check by decompressing and decrypting the data
+    extra_check: bool,
```
simonsan, line 341:
```suggestion
    integrity_check: bool,
```

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
             }
             None => self.key().encrypt_data(data)?,
         };
-        let id = hash(&data);
-        self.write_bytes(tpe, &id, false, data.into())
+        if self.extra_check {
```
simonsan, line 414: If we do it as a default, do we want to branch here, or would we rather branch when integrity_check is not set?

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
...
+            let check_data = self.decrypt_file(&data_encrypted)?;
```
simonsan, line 415: We might also want to implement a test on our side, like `data == decrypt(encrypt(data))` with different compression levels, just to be on the safe side for the future, that we catch possible regressions in that regard. I know, the library itself should actually do this and it's not our thing. But as we are relying here on this kind of existential feature of another crate, it might be good, to also secure the integrity on our side? 🤔 Not sure though.

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
...
+            .map_err(CryptBackendErrorKind::IntConversionFailed)?;
+        let (data_encrypted, uncompressed_length) = match self.zstd {
+            None => (self.key.encrypt_data(data)?, None),
+            // compress if requested
+            Some(level) => (
+                self.key.encrypt_data(&encode_all(data, level)?)?,
+                NonZeroU32::new(data_len),
+            ),
+        };
+        if self.extra_check {
```
simonsan, line 439: same here, if it's default, maybe we shouldn't branch here, but rather branch `if !self.integrity_check`?

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
...
+            let data_check =
+                self.read_encrypted_from_partial(&data_encrypted, uncompressed_length)?;
+            if data != data_check {
+                return Err(CryptBackendErrorKind::ExtraCheckFailed.into());
```
simonsan, line 443:
```suggestion
                return Err(CryptBackendErrorKind::IntegrityCheckFailed.into());
```

simonsan, line 443: If we are doing it twice, to keep it DRY, maybe we should factor that out to its own function and `inline` it?

Also: I'm reviewing online, and need to dive into the code late, but just as a remark: `read_encrypted_from_partial` is actually return unencrypted data here to compare?

crates/core/src/backend/decrypt.rs
```diff
@@ -319,6 +321,7 @@ pub trait DecryptWriteBackend: WriteBackend + Clone + 'static {
     }
 
     fn set_zstd(&mut self, zstd: Option<i32>);
```
simonsan, line 323:
```suggestion
    /// Sets the compression level to use for zstd.
    ///
    /// # Arguments
    ///
    /// * `zstd` - The compression level to use for zstd. TODO: What happens if this is None? What are defaults?
    fn set_zstd(&mut self, zstd: Option<i32>);
```

crates/core/src/backend/decrypt.rs
```diff
@@ -319,6 +321,7 @@ pub trait DecryptWriteBackend: WriteBackend + Clone + 'static {
...
+    fn set_extra_check(&mut self, extra_check: bool);
```
simonsan, line 324:
```suggestion

    /// Sets `integrity_check`, i.e. whether integrity checking the data after compressing/encrypting
    ///
    /// # Arguments
    ///
    /// * `integrity_check` - Internally sets a boolean for an integrity_check
    fn set_integrity_check(&mut self, integrity_check: bool);
```

crates/core/src/backend/decrypt.rs
```diff
@@ -409,6 +454,15 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
...
         self.zstd = zstd;
     }
+
+    /// Sets `extra_check`, i.e. whether to do an extra check after compressing/encrypting
+    ///
+    /// # Arguments
+    ///
+    /// * `extra_echeck` - The compression level to use for zstd.
+    fn set_extra_check(&mut self, extra_check: bool) {
+        self.extra_check = extra_check;
```
simonsan, line 464:
```suggestion
    /// Sets `integrity_check`, i.e. whether integrity checking the data after compressing/encrypting
    ///
    /// # Arguments
    ///
    /// * `integrity_check` - Internally sets a boolean for an integrity_check
    fn set_integrity_check(&mut self, integrity_check: bool) {
        self.integrity_check= integrity_check;
```

crates/core/src/backend/dry_run.rs
```diff
@@ -112,11 +112,24 @@ impl<BE: DecryptFullBackend> DecryptWriteBackend for DryRunBackend<BE> {
...
         if !self.dry_run {
             self.be.set_zstd(zstd);
         }
     }
+
+    fn set_extra_check(&mut self, extra_check: bool) {
+        if !self.dry_run {
+            self.be.set_extra_check(extra_check);
+        }
+    }
```
simonsan, line 132:
```suggestion
    fn set_integrity_check(&mut self, integrity_check: bool) {
        if !self.dry_run {
            self.be.set_integrity_check(integrity_check);
        }
    }
```

crates/core/src/commands/config.rs
```diff
@@ -158,6 +158,10 @@ pub struct ConfigOptions {
     /// tolerated. Default if not set: larger packfiles are always tolerated.
     #[cfg_attr(feature = "clap", clap(long, value_name = "PERCENT"))]
     pub set_max_packsize_tolerate_percent: Option<u32>,
+
+    /// Do an extra check by decompressing/decrypting all data before uploading to the repository
+    #[cfg_attr(feature = "clap", clap(long))]
+    pub set_extra_check: Option<bool>,
```
simonsan, line 164: Set as a default even here?

```suggestion
    /// Do an integrity_check by decompressing/decrypting all data before uploading to the repository
    #[cfg_attr(feature = "clap", clap(long, default = Some(True)))]
    pub set_integrity_check: Option<bool>,
```

crates/core/src/commands/config.rs
```diff
@@ -258,6 +262,8 @@ impl ConfigOptions {
             config.max_packsize_tolerate_percent = Some(percent);
         }
 
+        config.extra_check = self.set_extra_check;
```
simonsan, line 265:
```suggestion
        config.integrity_check = self.set_integrity_check;
```

crates/core/src/error.rs
```diff
@@ -569,6 +569,10 @@ pub enum CryptBackendErrorKind {
     EncryptingDataFailed,
     /// Compressing and appending data failed: `{0:?}`
     CopyEncodingDataFailed(std::io::Error),
+    /// conversion for integer failed: `{0:?}`
+    IntConversionFailed(#[from] TryFromIntError),
+    /// Extra check failed: After decrypting/decompressing the data doesn't match any longer!
+    ExtraCheckFailed,
```
simonsan, line 575:
```suggestion
    /// Integrity check failed: After decryption and decompression the data changed!
    IntegrityCheckFailed,
```

crates/core/src/repofile/configfile.rs
```diff
@@ -93,6 +93,9 @@ pub struct ConfigFile {
     ///
     /// If not set or set to `0` this is unlimited.
     pub max_packsize_tolerate_percent: Option<u32>,
+
+    /// Do an extra check by decompressing/decrypting all data before uploading to the repository
+    pub extra_check: Option<bool>,
```
simonsan, line 98:
```suggestion
    /// Do an integrity check by decompressing/decrypting all data before uploading to the repository
    pub integrity_check: Option<bool>,
```

crates/core/src/repofile/configfile.rs
```diff
@@ -146,6 +149,11 @@ impl ConfigFile {
         }
     }
 
+    /// Get wheter an extra check (decompressing/decrypting data before writing to the repository) should be performed.
+    pub fn extra_check(&self) -> bool {
+        self.extra_check.unwrap_or(true) // default is to do the extra check
+    }
```
simonsan, line 155:
```suggestion
    /// Whether to do an integrity check (decompressing/decrypting data before writing to the repository) should be performed.
    pub fn integrity_check(&self) -> bool {
        self.integrity_check.unwrap_or(true) // default is to do the extra check
    }
```

crates/core/src/repository.rs
```diff
@@ -632,8 +632,8 @@ impl<P, S> Repository<P, S> {
         }
 
         let mut dbe = DecryptBackend::new(self.be.clone(), key);
-        let zstd = config.zstd()?;
-        dbe.set_zstd(zstd);
+        dbe.set_zstd(config.zstd()?);
+        dbe.set_extra_check(config.extra_check());
```
simonsan, line 636:
```suggestion
        dbe.set_integrity_check(config.integrity_check());
```

simonsan (commented):
A few comments:
- would change the name to `integrity_check` because it feels more clear than `extra_check` in being descriptive in what it does.
- made sure at several places the default is being `true` rather than just in the config getter
- added a few doc comments
- added comments on branching, if it's the default

On another note: Not knowledgable about this, but maybe it is worth to explicitly check, if we took the right steps to make the compiler not optimize the check-out because it is smarter than us?

crates/core/src/backend/decrypt.rs
simonsan: Or actually turn it around, and say `self.no_integrity_check`?

crates/core/src/backend/decrypt.rs
aawsome: No - this is implemented here as defined in the restic repo specification.

crates/core/src/backend/decrypt.rs
aawsome: No - any test we find useful here, should instead be added to the library. And I think there are more competent people to decide on what tests are useful..

duracell:
Did you do a benchmark against real data? I'm just not sure if zstd may just skip non-compressible data.
Sorry, I'm not an expert on that field and don't know that much about the used lib, I only know that in some environments (zfs) this happens. I just want to be sure that the good performance is not a result of the random data and your choice, even if it's my proposal, is only a result of a edge case.

aawsome:
About the name: I don't like things like `--integrity-check` or `--no-integrity-check` because it may suggest that we don't do integrity check if not chosen. But in fact, a lot of integrity checks are done independently from this change. On the other hand, as already said: No, we don't double-check everything here but just compression+encryption. People shouldn't think that this option changes that much, it does only add a single extra check (which normally is/should be unneccessary) and not more.

duracell:
What's about the restic name "no-extra-verify"? I think this fits very well, also to your explanation. Or something else with "extra", "additional", "more"...?

aawsome:
> What's about the restic name "no-extra-verify"? I think this fits very well, also to your explanation. Or something else with "extra", "additional", "more"...?

Actually I like `extra-verify` most. As we anyhow have it in the config and users must only set it once per repo, it can be unset by `init --extra-verify=false` which is IMO is very clear and clean. We anyway have https://github.com/rustic-rs/rustic/issues/908 open which (when implemented) should be able to show that `extra-verify=true` if users didn't specify anything.

crates/core/src/repository.rs
simonsan: other name chosen: `extra-verfiy`

crates/core/src/repofile/configfile.rs
simonsan: other name chosen: `extra-verify`

simonsan:
I still have a tendency to like `--no-extra-verify` more, as a flag. Because it makes a bit more clear, that you are deactivating a default by setting that flag. Also using the same naming convention as restic here, makes it easier to transfer that knowledge of the flag to rustic, IMHO.

aawsome:
Negative names IMO tend to be worse UX. If you have `--no-extra-verify` and write `default: false` this is much more irritating instead of simply having `--extra-verify` and writing `default: true`...

crates/core/src/backend/decrypt.rs
```diff
@@ -356,8 +361,19 @@ impl<C: CryptoKey> DecryptBackend<C> {
             be,
             key,
             zstd: None,
+            extra_check: false,
```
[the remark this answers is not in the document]
aawsome: this is no argument.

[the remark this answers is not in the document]
aawsome: No. That we have it as default doesn't mean we want to set in always, just when we really want to write compressed/encrypted data to the repository. The DecryptBackend is also used in other edge cases.

aawsome:
> Did you do a benchmark against real data? I'm just not sure if zstd may just skip non-compressible data. Sorry, I'm not an expert on that field and don't know that much about the used lib, I only know that in some environments (zfs) this happens. I just want to be sure that the good performance is not a result of the random data and your choice, even if it's my proposal, is only a result of a edge case.

Sorry for the late reply: Yes, there were similar results for `/dev/null` (which only saves 1 blob and therefore has a very small overhead)  as well as with some real-life data. But this was all on the same hardware, could be different on other CPUs...

simonsan:
> Negative names IMO tend to be worse UX. If you have `--no-extra-verify` and write `default: false` this is much more irritating instead of simply having `--extra-verify` and writing `default: true`...
> 
> Or in a future `rustic config show`:
> 
> ```
> no-extra-verify: false
> ```

I think this could be labelled as intentionally worse UX, to make it harder to do the wrong things and easier to do the right ones. `extra-verify: false` sounds like it would be a totally reasonable thing to do, just another option kind of. As that might impact the integrity of a backup on higher compression levels, it should be standing out. I'm as well not sure, if `no-extra-verify` would be considered enough. I'm proposing the following:

We call it:
```
--skip-verification
    Skip default verification checks. NOT RECOMMENDED. This flag should only be used if you understand the potential risks and have a specific reason for bypassing verification.
```

And we log a warning to the console and even make it explicit to continue or just let it at the flag.

```
WARNING: Skipping default verification checks is NOT RECOMMENDED and can lead to data corruption or loss. Proceed with caution. Are you sure you want to continue? [y/N]:
```

duracell:
To add my comment about the naming, I liked the negative name for a command switch and the positive for the config, so no real preference for me. =/

> `WARNING: Skipping extra verification checks is NOT RECOMMENDED and can lead to data corruption or loss.`

I think this sounds too hard, as this option itself can cause data corruption. My suggestion would be:
`WARNUNG: Skipping extra verification checks is NOT RECOMMENDED as it hampers the detection of data corruption prior to uploading. (It will detected with a check --read-data).`
But not sure if this really sounds better 😅

aawsome:
> extra-verify: false sounds like it would be a totally reasonable thing to do, just another option kind of.

I think *this* is where we disagree. I do think that that is totally reasonable in rustic.

Actually I think it could add well be that we change the default in future if it poses any problems we haven't seen yet 

Maybe you should ask the guys from Facebook about their judgement regarding the fixed bug last year and your concerns about data integrity before we speculate if this is something which really is important...

simonsan:
I would see that more as a `Defense in depth` kind of mechanism, to not let rustic users be affected by possible bugs in the future. So in that regard it's not even as important how high the likelihood of something like that is to happen, as long as the likelihood is not 0. The verification helps to make sure, that data integrity of a backup is kept - right? (correct me if I'm wrong here)

Given that assumption, I do think it's totally reasonable:
- to make an extra verification step a default, even if performance of a backup would take a hit (I think the main feature of a backup software should be reliability, integrity, and correctness. Performance is a nice to have, IMHO. Do we agree here?)

- to make it less likely people accidentally deactivate that, but choosing clear names and giving a warning, that extra verification has been disabled

- to strive for correctness and reliability, by implementing more tests, improve error handling, etc.

aawsome:
(sorry - seems I deleted my prior post - was working on my mobile and thought I double-posted)

> So in that regard it's not even as important how high the likelihood of something like that is to happen, as long as the likelihood is not 0. The verification helps to make sure, that data integrity of a backup is kept - right? (correct me if I'm wrong here)

I think you are wrong:
- It is all about probabilities - and we don't (can) check for everything which can happen with probability >0: We use hashes to deduplicate. If we have a hash collision, we do loose data. This is with very low probability, but the probability is not zero. But we still don't check for hash collisions, because we also don't check for a lot of physical things which would also damage our data and which are much more likely, like: defect memory, defect CPUs, or the fact that an asteroid could hit the earth destroying all life and all computers. These are all things which are quite unlikely, but probability is >0 for all of these incidents. And, we also don't check for data loss of the storage used (for hard discs, this is *much*, *much* more likely than all of the above).
- Because of all the things we are not discussing here, we have to recall: This check does help to ensure data integrity, but it only does for the compression/encryption step and not for all other things. 

To give an example: When the actual repository files are saved to the storage, this may be done by sending the data over TCP to a storage provider. Alone this transmission will pass *a lot* of encoding/decoding steps where connections might even use zstd for some piece of the way. So, what does it help that we checked that *our* zstd doesn't corrupt data if some encoding algorithm (and it might even be zstd) used by some ISP finally corrupts it?

TL;DR: We do have to discuss probabilities when we add extra checks as we can *never* be sure about data integrity, so we can only improve probabilities. However, here we simply cannot judge probabilties of such a bug still being present (or being introduced in future), so the simple reason we could want to make the check default is that it is a) very cheap and b) might make people feel better ;-)

duracell:
> So, what does it help that we checked that _our_ zstd doesn't corrupt data if some encoding algorithm (and it might even be zstd) used by some ISP finally corrupts it?

That doesn't help, but it reduces the chances by itself. And depending on the backend you're using this will be handled by this backend. If you using rclone with backblaze you need to give them the sha1(?) hash of the file you're uploading and they compare it on their site after the received the file. So the example was not the best. But of course you're right, 100% protection is not possible. But I think for this problem we have a good agreement of that's unlikely, but far from impossible, because a similar problem already happened twice in another lib and once in the used lib.

> TL;DR: We do have to discuss probabilities when we add extra checks as we can _never_ be sure about data integrity, so we can only improve probabilities. However, here we simply cannot judge probabilties of such a bug still being present (or being introduced in future), so the simple reason we could want to make the check default is that it is a) very cheap and b) might make people feel better ;-)

That's true :)

simonsan:
>  So, what does it help that we checked that our zstd doesn't corrupt data if some encoding algorithm (and it might even be zstd) used by some ISP finally corrupts it?

Some thoughts:

1. There will always be things out of our control in such a complex system, but we **should control what we can** and make sure the highest possible data integrity is ensured on our side.
1. This assures ourselves and our users, that the data corruption didn't happen on our side. (Layers of defence)
1. Thus making it easier, to diagnose/debug the problem due to it being isolated to a different part of the processing chain. (Problem isolation)
1. For a backup tool, reliability, and trust are foundational. Trust needs to be constantly build and maintained.
1. Adhering to industry standards and compliance requirements often introduces the need for rigorous testing and validation of data integrity measures.
1. I understand, that it's kind of an external factor that we are talking about here. But we are pulling this in as a library, so as a user and creator of a backup tool to others, we also need to make sure, that data integrity is kept, when it leaves our side.

> But of course you're right, 100% protection is not possible. But I think for this problem we have a good agreement of that's unlikely, but far from impossible, because a similar problem already happened twice in another lib and once in the used lib.

I would even argue, that 100% protection is not possible, but we can **AND SHOULD** always do things to get in that direction and to lower/control the impact of things becoming corrupted - error detection and error correction. So even if there are endless possibilities, how something can become erroneous, it wouldn't be necessarily a reason to NOT do something or shift the responsibility.

crates/core/src/backend/decrypt.rs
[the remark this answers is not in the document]
aawsome: IMO this can be added to the docu, but here it is not useful.

crates/core/src/backend/decrypt.rs
```diff
@@ -395,12 +411,41 @@ impl<C: CryptoKey> DecryptWriteBackend for DecryptBackend<C> {
...
+                return Err(CryptBackendErrorKind::ExtraCheckFailed.into());
+            }
+        }
+        let id = hash(&data_encrypted);
+        self.write_bytes(tpe, &id, false, data_encrypted.into())
             .map_err(RusticErrorKind::Backend)?;
         Ok(id)
     }
 
+    fn process_data(&self, data: &[u8]) -> RusticResult<(Vec<u8>, u32, Option<NonZeroU32>)> {
```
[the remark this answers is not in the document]
aawsome: it is documented in the trait.

crates/core/src/commands/repair/snapshots.rs
```diff
@@ -180,7 +180,7 @@ impl RepairSnapshotsOptions {
     /// # Returns
     ///
     /// A tuple containing the change status and the id of the repaired tree
-    fn repair_tree<BE: DecryptWriteBackend>(
+    fn repair_tree<BE: DecryptFullBackend>(
```
aawsome, line 183: this is no longer needed -> undo that change!

crates/core/src/backend/decrypt.rs
aawsome: actually I think we should prefer clean code to premature optimization. (I doubt that branch prediction will make any measurable difference here)

crates/core/src/commands/config.rs
aawsome: I think we should overwork this when tackling https://github.com/rustic-rs/rustic/issues/908

crates/core/src/backend/decrypt.rs
aawsome: sorry, don't see the repetition. And yes, it does return decrypted data.

aawsome:
@simonsan Is there anything open here? If not I would like to merge this PR.

PR #155, opened 2024-02-07
feat(opendal): Add option connections

aawsome:
CC: https://github.com/rustic-rs/rustic/issues/981


crates/backend/src/opendal.rs
```diff
@@ -45,19 +45,30 @@ impl OpenDALBackend {
...
+            .transpose()?;
 
         let schema = Scheme::from_str(path.as_ref())?;
+        let mut operator = Operator::via_map(schema, options)?
+            .layer(RetryLayer::new().with_max_times(max_retries).with_jitter());
+
+        if let Some(connections) = connections {
+            operator = operator.layer(ConcurrentLimitLayer::new(connections));
+        }
+
```
simonsan, line 65: Would it make sense to have reasonable defaults here? E.g. the same as max parallelism?

simonsan: Also, would it make sense, to use the same value for the thread pool in `rustic` then or should that be an extra option?

aawsome: I don't know about what defaults are reasonable, TBH. There is also no way to get opendals defaults, the only way is to not use the ConcurrencyLimit layer...

About the parallelism used in rustic, I would postpone this to a performance-targeting refactor...

aawsome: But anyhow, why should we put effort into setting the opendal default ourselves - if we can just use it by not adding the layer. 
So, actually I would just let it be as it is. And most users most likely won't specify a `connections` options. They only will if they run into trouble without.

simonsan: Yeah, it was more in the direction, if there is something we could do about it. Problem is, I'm affected by this, as I use Hetzner Storage Box which is consistently failing for these kinda things. :/ But otherwise I'm fine with it, it's really just the backup stuff with rclone currently. Where we don't have any options. So I hope to switch to opendal soon.

simonsan (approved):

PR #156, opened 2024-02-07
feat: add FuseFs

aawsome:
needed to implement a `mount` command

TODO:
- [ ] decide platforms to support
- [ ] Think about how to setup the `fuse` feature (unfortunately, there are no target-specific dependencies, see https://github.com/rust-lang/cargo/issues/1197)
- [ ] Add needed dependencies to CI-Pipeline for chosen platforms

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -39,6 +39,7 @@ default = []
 cli = ["merge", "clap"]
 merge = ["dep:merge"]
 clap = ["dep:clap"]
+fuse = ["dep:fuse_mt"]
 webdav = ["dep:dav-server", "dep:futures"]
 
 [package.metadata.docs.rs]
@@ -97,6 +98,7 @@ merge = { version = "0.1.0", optional = true }
 
 # vfs support
 dav-server = { version = "0.5.8", default-features = false, optional = true }
+fuse_mt = { version = "0.6", optional = true }
 futures = { version = "0.3", optional = true }
 runtime-format = "0.1.3"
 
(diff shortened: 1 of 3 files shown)


aawsome:
https://github.com/rustic-rs/rustic/pull/973 now implements `FuseFS` in `rustic-rs` which I think is the better place.

simonsan:
> [rustic-rs/rustic#973](https://github.com/rustic-rs/rustic/pull/973) now implements `FuseFS` in `rustic-rs` which I think is the better place.

Just to understand, why do you think `rustic-rs` is the better place?

Related: https://github.com/rustic-rs/rustic_core/issues/332

aawsome:
> > [rustic-rs/rustic#973](https://github.com/rustic-rs/rustic/pull/973) now implements `FuseFS` in `rustic-rs` which I think is the better place.
> 
> Just to understand, why do you think `rustic-rs` is the better place?

`FuseFS` is only needed to offer a fuse filesystem to be used by `fuse_mt`. This adds dependencies into rustic_core but is actually not a functionality very related to the core of rustic.

Moreover, the code using `FuseFS` also need the identical dependency itself. And we could ask us, why we support `fuse_mt` but not `fuser`directly - or `fuse3` (which we may use in future). By only offering basic functionality in `vfs` but not a coded "mapping" we make rustic_core more lightweight without loosing any functionality.

N.B: It is actually similar with the `merge` and `cli` features - those are in principle also nothing we need in rustic_core. But here it having the traits not implemented would basically lead to very much boilerplate code on the caller sider - they would basically copy the structs while rustic_core would not get much more lightweight. For `FuseFS`,  not implementing it in rustic_core reduces dependencies and code complexity in rustic_core.

PR #163, opened 2024-02-16
feat: Add Repository::lock_repo

aawsome:
Adds `Repository::lock_repo` which locks all repository files. Also the backend trait is extended to support locking and a `lock_command` has been added to `RepositoryOptions` which turns any backend into a backend capable of locking.

see https://github.com/rustic-rs/rustic/discussions/1050

Missing points:
- [ ] Add unit tests

### DIFF
--- a/crates/core/src/backend.rs
+++ b/crates/core/src/backend.rs
@@ -358,7 +358,15 @@ pub trait WriteBackend: ReadBackend {
     /// If the file could not be read.
     fn lock(&self, tpe: FileType, id: &Id, until: Option<DateTime<Local>>) -> RusticResult<()> {
         debug!("no locking implemented. {tpe:?}, {id}, {until:?}");
-        Ok(())
+
+        if self.can_lock() {
+            unimplemented!("Using default implementation. No locking implemented in backend.");
+        } else {
+            Err(RusticError::new(
+                ErrorKind::Backend,
+                "No locking configured on backend.",
+            ))
+        }
     }
 }
 
--- a/crates/core/src/backend.rs
+++ b/crates/core/src/backend.rs
@@ -6,6 +6,7 @@ pub(crate) mod dry_run;
 pub(crate) mod hotcold;
 pub(crate) mod ignore;
 pub(crate) mod local_destination;
+pub(crate) mod lock;
 pub(crate) mod node;
 pub(crate) mod stdin;
 pub(crate) mod warm_up;
@@ -13,8 +14,9 @@ pub(crate) mod warm_up;
 use std::{io::Read, ops::Deref, path::PathBuf, sync::Arc};
 
 use bytes::Bytes;
+use chrono::{DateTime, Local};
 use enum_map::Enum;
-use log::trace;
+use log::{debug, trace};
 
 #[cfg(test)]
 use mockall::mock;
@@ -337,6 +339,27 @@ pub trait WriteBackend: ReadBackend {
     ///
     /// The result of the removal.
     fn remove(&self, tpe: FileType, id: &Id, cacheable: bool) -> RusticResult<()>;
+
+    /// Specify if the backend is able to lock files
+    fn can_lock(&self) -> bool {
+        false
+    }
+
+    /// Lock the given file.
+    ///
+    /// # Arguments
+    ///
+    /// * `tpe` - The type of the file.
+    /// * `id` - The id of the file.
+    /// * `until` - The date until when to lock. May be `None` which usually specifies a unlimited lock
+    ///
+    /// # Errors
+    ///
+    /// If the file could not be read.
+    fn lock(&self, tpe: FileType, id: &Id, until: Option<DateTime<Local>>) -> RusticResult<()> {
+        debug!("no locking implemented. {tpe:?}, {id}, {until:?}");
+        Ok(())
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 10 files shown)


crates/core/src/backend.rs
```diff
@@ -337,6 +339,27 @@ pub trait WriteBackend: ReadBackend {
...
+    /// * `id` - The id of the file.
+    /// * `until` - The date until when to lock. May be `None` which usually specifies a unlimited lock
+    ///
+    /// # Errors
+    ///
+    /// If the file could not be read.
+    fn lock(&self, tpe: FileType, id: &Id, until: Option<DateTime<Local>>) -> Result<()> {
+        debug!("no locking implemented. {tpe:?}, {id}, {until:?}");
+        Ok(())
+    }
```
simonsan, line 362: I would check here for `::can_lock()` and if that is true and `::lock()` is being called in the default implementation, I would `panic!()` or `not_implemented!()` as otherwise someone could actually think, they can use this and won't reimplement it. It should definitely not silently continue here.

crates/core/src/commands.rs
```diff
@@ -12,6 +12,7 @@ pub mod dump;
 pub mod forget;
 pub mod init;
 pub mod key;
+pub mod lock;
```
simonsan, line 15: We don't expose anything via our API? There are no changes to `lib.rs`.

crates/core/src/backend/lock.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+fn path(tpe: FileType, id: &Id) -> String {
+    let hex_id = id.to_hex();
+    match tpe {
+        FileType::Config => "config".into(),
+        FileType::Pack => format!("data/{}/{}", &hex_id[0..2], &*hex_id),
+        _ => format!("{}/{}", tpe.dirname(), &*hex_id),
+    }
+}
```
simonsan, line 78: Is there a better name for this helper function? Like `path_to_id_from_file_type` or something?

crates/core/src/backend/lock.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn remove(&self, tpe: FileType, id: &Id, cacheable: bool) -> Result<()> {
+        self.be.remove(tpe, id, cacheable)
+    }
+
+    fn can_lock(&self) -> bool {
+        true
+    }
+
+    fn lock(&self, tpe: FileType, id: &Id, until: Option<DateTime<Local>>) -> Result<()> {
```
simonsan, line 97: `::lock()` should never be callable, if `::can_lock()` isn't returning `true`, I think?

crates/core/src/commands/lock.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+pub fn lock_all_files<P: ProgressBars, S, ID: RepoId + std::fmt::Debug>(
+    repo: &Repository<P, S>,
+    until: Option<DateTime<Local>>,
+) -> RusticResult<()> {
+    if !repo.be.can_lock() {
+        return Err(CommandErrorKind::NoLockingConfigured.into());
+    }
+
```
simonsan, line 39: I see we check that here, but I think this should be checked within the `::lock()` itself, so people using it, can't forget to call it.

crates/core/src/commands/lock.rs
```diff
@@ -0,0 +1,75 @@
...
+        .progress_counter(format!("locking {:?} files..", ID::TYPE));
+    p.set_length(ids.len().try_into().unwrap());
+    let backend = &repo.be;
+    pool.in_place_scope(|scope| {
+        for id in ids {
+            scope.spawn(move |_| {
+                if let Err(e) = backend.lock(ID::TYPE, id, until) {
+                    // FIXME: Use error handling
+                    error!("lock failed for {:?} {id:?}. {e}", ID::TYPE);
+                };
```
simonsan, line 68: Is it actually enough to just log an error here? If something like that `lock` can't be acquired, I would throw an error actually, so the command is not being run, right? The expectation on a function, that is called `lock_files` should be that is errors out, when it can't log the files, and not just logs an error I feel.

simonsan (changes requested):

crates/core/src/backend.rs
simonsan: Especially because it doesn't implement the expected behaviour.

crates/core/src/backend/lock.rs
simonsan: A type state design might be better suited here

simonsan:
I pushed some of my proposed changes in a commit, to better explain what I mean. I'm not so sure about the design between `can_lock()` and `lock()`, that needs to check itself, if `can_lock()` is `true` before running. This makes probably some room for bugs, if someone forgets to check it before running `lock()` - so not sure about the API design in that case.

crates/core/src/commands.rs
aawsome: We only expose new things on `RepositoryOptions` and a new method on `Repository`..

crates/core/src/backend/lock.rs
aawsome: But we do have a problem with the type state pattern here: The Repository traits need to be object save as we do create dynamic trait objects here. 
Do you know a way to implement a type state pattern in combination with object safety?

crates/core/src/commands/lock.rs
aawsome: This is the same error handling discussion as in many places: The best would here to log an error, continue and collect the errors and at the end fail with the list of errors...

simonsan: Yeah :/ But keeping it running and not failing early also opens room up for possible bugs I think that could affect repository integrity? (in general, not regarding the locking discussion).

When you say `at the end`, what do you mean by that. Just in the end of the program's run? Or at the end of the function?

crates/core/src/backend/lock.rs
simonsan: I need to look into it, but maybe https://crates.io/crates/enum_dispatch can help. 🤔

aawsome:
rebased this PR to current master

PR #170, opened 2024-02-25
feat: Hide plain text password from help text

Granddave:
When the password is provided via the environment variable like `RUSTIC_PASSWORD=hunter2`, it would be visible in the help text:

```
--password <PASSWORD>     Password of the repository [env: RUSTIC_PASSWORD=hunter2]
```

To avoid exposing any sensitive information this commit hides the password from the help text:

```
--password <PASSWORD>     Password of the repository [env: RUSTIC_PASSWORD]
```


aawsome (approved):
Thanks @Granddave for the PR - good catch!
LGTM

PR #175, opened 2024-03-05
test: add backup integration tests using snapshots

aawsome:
This PR adds some integration tests (mostly inspired by restic's integration tests). We use `insta` for "snapshotting" internal state and then compare against it depending on which operating system we are on.


crates/core/tests/integration.rs
```diff
@@ -0,0 +1,71 @@
...
+
+fn set_up_tar_fixture(path: impl AsRef<Path>) -> Result<TempDir> {
+    let dir = tempdir()?;
+    let path = Path::new("tests/testdata").join(path);
+    let tar_gz = File::open(path)?;
+    let tar = GzDecoder::new(tar_gz);
+    let mut archive = Archive::new(tar);
+    archive.unpack(&dir)?;
+    Ok(dir)
+}
```
simonsan, line 32: https://crates.io/crates/rstest fixtures are also nice for this, it works via DI

simonsan (dismissed):

simonsan:
Wanted to merge it, to test out the new merge queue CI, but maybe you also want to have a look before @aawsome? Especially if the numbers are plausible.

I reworked the integration test, so we have `ron` snapshots, because redactions were not possible in debug snapshots. But I think redactions can become useful going forward, also because the code is more structured. I passed in everything via dependency injection due to rstest's fixtures. I think that also gives a better overview, of what's actually being used in the function (after getting used to it). 

In addition: I factored out the redaction settings to their own fixtures as well, to make them reusable and easier to administer. I also wrote some comments on the top of the integration test file for some explanations. 

I think it's a good foundation for future integration tests. Furthermore, I do think we can add a few more, like these. For example, making sure, that backing up a directory and then a second time actually changes the `files_modified` and `files_new` fields, etc. But for now I think it's fine, and we should test more fine grainer (e.g. unit testing for `PathList` and others) to make sure, the basics are right on each plattform.

aawsome (commented):
LGTM! Thanks a lot @simonsan for overworking!

simonsan (approved):

PR #176, opened 2024-03-09
feat!: move clippy lints to cargo manifest and fix upcoming issues all over the workspace

### LINKED ISSUE #59: Move lints from `lib.rs` to `Cargo.toml`
Based on this new feature: https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

It would increase the MSRV though, so not really sure for now, maybe we wait for another stable release or two?

simonsan:
Fixes #59


crates/core/src/blob/packer.rs
```diff
@@ -91,6 +91,8 @@ impl PackSizer {
 
     /// Computes the size of the pack file.
     #[must_use]
+    // TODO!: IMPORTANT: Evaluate deeply if this is really fine and what we want
```
simonsan, line 94: This needs further evaluation

crates/core/src/cdc/polynom.rs
```diff
@@ -12,6 +12,8 @@ pub(crate) type Polynom64 = u64;
 
 impl Polynom for Polynom64 {
     /// The degree of the polynom.
+    // TODO!: IMPORTANT: Evaluate deeply if this is really fine and what we want
+    #[allow(clippy::cast_possible_wrap)]
     fn degree(&self) -> i32 {
         63 - self.leading_zeros() as i32
     }
```
simonsan, line 19: This needs further evaluation

crates/core/src/blob/packer.rs
aawsome: The cast actually shouldn't pose any problems.
`current_size` is `u64`, the maximum value is `2^64-1`.
`isqrt(2^64-1) = 2^32-1` which fits into a `u32`.

crates/core/src/commands/prune.rs
```diff
@@ -754,6 +755,8 @@ impl PrunePlan {
     /// # Errors
     ///
     // TODO: add errors!
+    #[allow(clippy::too_many_lines)]
+    #[allow(clippy::unnecessary_wraps)]
```
aawsome, line 759: actually this lint is right: This method is never returning an error.

crates/core/src/commands/restore.rs
```diff
@@ -696,9 +698,12 @@ impl RestorePlan {
             };
             let length = bl.data_length();
 
+            let usize_lenth =
+                usize::try_from(length).map_err(CommandErrorKind::ConversionFromIntFailed)?;
```
aawsome, line 702: This will never fail unless we are on a 16 bit machine. `length` is basically a `u32` converted into a `u64`. 
But maybe just keep it, however maybe rename into `usize_length`?

crates/core/src/repository.rs
```diff
@@ -286,7 +286,7 @@ impl Repository<NoProgressBars, ()> {
     /// * [`BackendAccessErrorKind::BackendLoadError`] - If the specified backend cannot be loaded, e.g. is not supported
     ///
     /// [`BackendAccessErrorKind::BackendLoadError`]: crate::error::BackendAccessErrorKind::BackendLoadError
-    pub fn new(opts: &RepositoryOptions, backends: RepositoryBackends) -> RusticResult<Self> {
+    pub fn new(opts: &RepositoryOptions, backends: &RepositoryBackends) -> RusticResult<Self> {
```
aawsome, line 289: Is this necessary? This breaks the API...

simonsan: It's not necessary, but we don't use it, so essentially the one before was unnecessary. I know it breaks the API, but maybe it's good to do that now, also the next release version is 0.3.0, so we essentially are already having API breaking changes, so that should be ok.

crates/core/src/commands/prune.rs
simonsan: Yes, it is. Currently. The problem is, I want to remove these things only, when we have taken care about the unwraps/expects. There might be a lot of API breaking changes coming there and maybe we will need a result here anyway. So for now this should be ok.

aawsome (approved):
LGTM

PR #188, opened 2024-03-11
fix(rclone): Use semver for version checking

### LINKED ISSUE #6: Explicitly set a flag to allow rclone < 1.52.2 to connect without authentication
```
// for rclone < 1.52.2 setting user/password via env variable doesn't work. This means
// we are setting up an rclone without authentication which is a security issue!
// (however, it still works, so we give a warning)
```
in `RCloneBackend::new()`

I feel like this should be made explicit by introducing a flag to agree to and otherwise error.

aawsome:
rclone version checking required a version in form "major.minor.patch". Now it supports all semver-compatible versions.

closes https://github.com/rustic-rs/rustic/issues/1097
fixes #6

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ rustic_core = { path = "crates/core" }
 simplelog = "0.12.2"
 
 # dev-dependencies
+rstest = "0.18.2"
 tempfile = "3.10.1"
 
 # see: https://nnethercote.github.io/perf-book/build-configuration.html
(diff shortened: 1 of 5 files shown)


simonsan (approved):

simonsan (approved):

simonsan:
Looks good to me now, wanna check as well @aawsome ?

aawsome (commented):
LGTM

simonsan:
Was merge conflict in merge queue, removed from queue and will fix it after merge of the other one

simonsan (approved):

PR #205, opened 2024-03-15
fix: Add warning about unsorted files and sort where neccessary

aawsome:
When a list of files is read from the backend, it comes randomly sorted because the read is parallelized.

This PR adds warnings in the docu for affected methods and orders the output of `SnapshotFile::from_ids`.

closes https://github.com/rustic-rs/rustic/issues/1103


simonsan:
Let's merge this after #202, ok? So we don't have to rebase the big PR. 👍🏽

PR #212, opened 2024-04-25
feat: add search methods to Repository

aawsome:
Adds `Repository` methods to find a given path in multiple trees.

`find_nodes_from_path`: Searches for an explicitly given path
`find_matching_nodes`: Searches using an arbitrary matching criterion


aawsome (commented):
LGTM

PR #213, opened 2024-04-25
fix(paths): Handle paths starting with "." correctly

aawsome:
See https://github.com/rustic-rs/rustic/issues/1135


aawsome (commented):
LGTM

PR #215, opened 2024-05-01
fix(errors): Show filenames in error message coming from ignore source

aawsome:
see https://github.com/rustic-rs/rustic/issues/1122

### DIFF
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -578,16 +578,17 @@ pub enum CryptBackendErrorKind {
 pub enum IgnoreErrorKind {
     /// generic Ignore error: `{0:?}`
     GenericError(#[from] ignore::Error),
-    /// Unable to open file: {0:?}
-    UnableToOpenFile(std::io::Error),
-    /// `{0:?}`
-    #[error(transparent)]
-    FromIoError(#[from] std::io::Error),
+    /// Error reading glob file {file:?}: {err:?}
+    ErrorGlob { file: PathBuf, err: std::io::Error },
+    /// Unable to open file {file:?}: {err:?}
+    UnableToOpenFile { file: PathBuf, err: std::io::Error },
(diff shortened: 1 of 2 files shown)


crates/core/src/error.rs
```diff
@@ -578,16 +578,17 @@ pub enum CryptBackendErrorKind {
...
-    #[error(transparent)]
-    FromIoError(#[from] std::io::Error),
+    /// Error reading glob file {file:?}: {err:?}
+    ErrorGlob { file: PathBuf, err: std::io::Error },
+    /// Unable to open file {file:?}: {err:?}
+    UnableToOpenFile { file: PathBuf, err: std::io::Error },
+    /// Error getting xattrs for {path:?}: {err:?}
+    ErrorXattr { path: PathBuf, err: std::io::Error },
+    /// Error reading link target for {path:?}: {err:?}
+    ErrorLink { path: PathBuf, err: std::io::Error },
```
simonsan, line 588: Please don't merge these PRs as fast as you do. Between opening and merging were 20 Minutes ... for no apparent reason? This is not a high priority hotfix or anything. So I really don't understand why this has to be merged without any review. Otherwise, I could have brought up the following:

`err` should be `source` here as that helps identifying the underlying error.

The Error trait's source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.

The #[from] attribute always implies that the same field is #[source], so you don't ever need to specify both attributes.

Any error type that implements std::error::Error or dereferences to dyn std::error::Error will work as a source.

```rust
#[derive(Error, Debug)]
pub struct MyError {
    msg: String,
    #[source]  // optional if field name is `source`
    source: anyhow::Error,
}
```

src.: https://crates.io/crates/thiserror

aawsome: We can add this with a follow-up PR. There have been already other errors in `error.rs` where the same applies.

PR #216, opened 2024-05-01
feat(backends): Add throttle option to opendal backend

aawsome:


crates/backend/src/opendal.rs
```diff
@@ -34,6 +35,30 @@ fn runtime() -> &'static Runtime {
...
+            .map(|s| ByteSize::from_str(s).map_err(|err| anyhow!("Error: {err}")))
+            .map(|b| -> Result<u32> { Ok(b?.as_u64().try_into()?) });
+        let bandwidth = values
+            .next()
+            .ok_or_else(|| anyhow!("no bandwidth given"))??;
+        let burst = values.next().ok_or_else(|| anyhow!("no burst given"))??;
+        Ok(Self { bandwidth, burst })
+    }
+}
+
```
simonsan, line 61: Can we implement a unit test for this, please? I think it makes sense to have that whenever we parse values from user input, to make sure it works as expected. Also, we add new code. The test coverage doesn't throw a warning here, because I took rustic_backend out of the reporting, due to complexity to test foreign code, but if we work with parsing and this may actually affect the user, it would be good, if this has test cases.

simonsan (changes requested):

crates/backend/src/opendal.rs
aawsome: unit tests are added.

simonsan (approved):

PR #220, opened 2024-05-03
fix clippy lints

aawsome:


simonsan:
Looks good, the other errors seem like a thing within `displaydoc`, using backticks there would break the display of the error messages. I can investigate, if there are fixes coming in displaydoc for that.

simonsan:
Asked upstream: https://github.com/yaahc/displaydoc/issues/48

crates/core/src/error.rs
```diff
@@ -1,5 +1,6 @@
 //! Error types and Result module.
 
+#![allow(clippy::doc_markdown)]
```
simonsan, line 3:
```suggestion
// Remove when 'displaydoc' has fixed/recommended further treatment upstream: https://github.com/yaahc/displaydoc/issues/48
#![allow(clippy::doc_markdown)]
```

simonsan (approved):

PR #224, opened 2024-05-06
fix(commands)!: return error if check fails

### LINKED ISSUE #222: Improve error handling in `check`
First, the most important thing is that `prune` (being the only command which removes pack or index files) doesn't leave the repository in an unsane condition. This is why `check` is called here. (to admit, check currently does not produce an error but only logs an `error!` - I'm about to improve this next).

_Originally posted by @aawsome in https://github.com/rustic-rs/rustic_core/pull/221#discussion_r1590161138_

aawsome:
closes #222


crates/core/src/commands/check.rs
```diff
@@ -37,7 +38,41 @@ pub struct CheckOptions {
     pub read_data: bool,
 }
 
-impl CheckOptions {
+#[allow(clippy::enum_glob_use)]
+use CheckError::*;
```
simonsan, line 42: I think glob imports are bad because they hinder readability and confuse. They are usually being frowned upon, so I think they should be removed here as well, and just the normal enum syntax should be used. I removed it all over the crate, when I factored out the library actually.

simonsan (changes requested):

crates/core/src/commands/check.rs
simonsan: It's also hard to review online, by the way. So I didn't go further still, will do it after the change, so I know where the enum is being used. 😬

aawsome: Ok, I removed the enum glob import...

crates/core/src/error.rs
```diff
@@ -790,3 +797,83 @@ where
         Self(RusticErrorKind::from(value))
     }
 }
+
+#[non_exhaustive]
+#[derive(Error, Debug, Display)]
+pub enum CheckError {
+    /// error reading pack {id} : {err}
+    ErrorReadingPack { id: Id, err: RusticError },
```
simonsan, line 805:
```suggestion
    ErrorReadingPack { id: Id, err: Box<RusticError> },
```

simonsan: I assume it makes sense to also box this one here, as the RusticError became quite large over time

crates/core/src/lib.rs
```diff
@@ -119,6 +119,8 @@ pub(crate) mod repository;
 /// Virtual File System support - allows to act on the repository like on a file system
 pub mod vfs;
 
+pub use log::Level as LogLevel;
+
```
simonsan, line 123: Is that an intended re-export? I assume we might want to use `tracing` in the future (especially when refactoring to `actors` or refactoring the threaded stuff), not sure if we want to have this as part of our API? 🤔

crates/core/src/commands/check.rs
```diff
@@ -37,7 +38,38 @@ pub struct CheckOptions {
     pub read_data: bool,
 }
 
-impl CheckOptions {
+#[derive(Debug)]
+pub struct CheckResults {
+    pub errors: Vec<(LogLevel, CheckError)>,
```
simonsan, line 43: Ah, I see now, this is the reason it is part of the public API, I guess. Hmm, maybe we can wrap it in our own LogLevel type or something, so we could essentially just swap log out against tracing for example under the hood?

crates/core/src/lib.rs
aawsome: is now changed by an own enum.

simonsan:
I'll look through it tomorrow. 👍🏽

crates/core/src/commands/check.rs
```diff
@@ -37,7 +37,54 @@ pub struct CheckOptions {
...
 }
 
-impl CheckOptions {
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+/// `CheckErrorLevel` describes severity levels of problems identified by check.
+pub enum CheckErrorLevel {
+    /// A warning: Something is strange and should be corrected, but repository integrity is not affected.
+    Warn,
+    /// An erro: Something in the repository is not as it should be.
```
simonsan, line 46:
```suggestion
    /// An error: Something in the repository is not as it should be.
```

crates/core/src/commands/check.rs
```diff
@@ -37,7 +37,54 @@ pub struct CheckOptions {
...
+    Error,
+}
+
+#[derive(Debug)]
+/// `CheckResults` is a list of errors encountered during the check.
+pub struct CheckResults {
+    /// The list of errors with severity level.
+    pub errors: Vec<(CheckErrorLevel, CheckError)>,
+}
```
simonsan, line 55: Wondering if we could replace these with: https://crates.io/crates/lazy_errors

Didn't use it, but it sounded interesting on Reddit last week: https://www.reddit.com/r/rust/comments/1ddi1nb/i_published_my_first_crate_lazy_errors/

crates/core/src/commands/check.rs
```diff
@@ -37,7 +37,54 @@ pub struct CheckOptions {
     pub read_data: bool,
 }
 
-impl CheckOptions {
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+/// `CheckErrorLevel` describes severity levels of problems identified by check.
+pub enum CheckErrorLevel {
```
simonsan, line 43: I would probably stick to the `kind` naming scheme we have in rustic_core, so something like `CheckErrorKind`.

crates/core/src/commands/check.rs
```diff
@@ -52,13 +99,17 @@ impl CheckOptions {
     /// # Errors
     ///
     /// If the repository is corrupted
-    pub(crate) fn run<P: ProgressBars, S: Open>(self, repo: &Repository<P, S>) -> RusticResult<()> {
+    pub(crate) fn check<P: ProgressBars, S: Open>(
+        &self,
+        opts: CheckOptions,
```
simonsan, line 104: I like that much better, it's a good usage of structs, to pass these inside like this, so it's also easy to `impl Default on CheckOptions` 👍🏽

crates/core/src/commands/check.rs
```diff
@@ -71,18 +122,18 @@ impl CheckOptions {
 
                     let p = pb.progress_bytes(format!("checking {file_type:?} in cache..."));
                     // TODO: Make concurrency (20) customizable
-                    check_cache_files(20, cache, raw_be, file_type, &p)?;
+                    self.check_cache_files(20, cache, raw_be, file_type, &p)?;
```
simonsan, line 125: Not sure, if that could be just a field in `CheckOptions`? or maybe another struct `ParallelismOptions` composed into `CheckOptions` that we can use in rustic_core.

crates/core/src/commands/check.rs
```diff
@@ -92,10 +143,10 @@ impl CheckOptions {
             }
             p.finish();
 
-            if !self.trust_cache {
+            if !opts.trust_cache {
                 let p = pb.progress_bytes("checking packs in cache...");
                 // TODO: Make concurrency (5) customizable
-                check_cache_files(5, cache, raw_be, FileType::Pack, &p)?;
+                self.check_cache_files(5, cache, raw_be, FileType::Pack, &p)?;
```
simonsan, line 149: Same here, maybe `ParallelismOptions` could be useful here as well?

crates/core/src/commands/check.rs
```diff
@@ -125,391 +176,458 @@ impl CheckOptions {
                 .for_each(|pack| {
                     let id = pack.id;
                     let data = be.read_full(FileType::Pack, &id).unwrap();
-                    match check_pack(be, pack, data, &p) {
+                    match self.check_pack(be, pack, data, &p) {
                         Ok(()) => {}
-                        Err(err) => error!("Error reading pack {id} : {err}",),
+                        Err(err) => self.add_error(CheckError::ErrorReadingPack {
+                            id,
+                            err: Box::new(err),
```
simonsan, line 183: would call it `source` so it's more clear

crates/core/src/commands/check.rs
```diff
@@ -125,391 +176,458 @@ impl CheckOptions {
...
-                error!("Type: {file_type:?}, Id: {id}: hot size: {size_hot}, actual size: {size}");
-            }
-            _ => {} //everything ok
+    fn add_error(&self, err: CheckError) {
+        if self.log {
+            error!("{err}");
         }
+        self.errors
+            .lock()
+            .unwrap()
```
simonsan, line 198: would at least prefer an `expect` here

crates/core/src/error.rs
```diff
@@ -790,3 +797,83 @@ where
         Self(RusticErrorKind::from(value))
     }
 }
+
+#[non_exhaustive]
+#[derive(Error, Debug, Display)]
+pub enum CheckError {
```
simonsan, line 803: Shouldn't that also be convertable to RusticError?

crates/core/src/commands/check.rs
simonsan: Not relevant, misread it

crates/core/src/commands/check.rs
simonsan: That looks still relevant to me, lazy_errors looks actually really good for collecting multiple errors like CheckResults does but looks more ergonomic to me at first look.

simonsan:
After looking through it, I understand better now, what I found confusing about it, back in May. It's really because it introduces a special way to handle errors only for `check`, although we want to have a unified way of handling errors. And apply that throughout `rustic_core`, which makes it easier to maintain, than having a special way to handle `check errors`.

The `CheckError` in `errors.rs` is really nice though, and this is probably the approach, we should take all over the code base. Then we might construct them in-place with all the information we need and create a well-formed error message towards the user from it.

For collecting the errors in a method/function we could use `lazy_errors` or change the signature to something like either:
- `Result<T, Vec<RusticError>>`
- ```rust
  struct ResultWithErrors<T> {
      value: Option<T>,
      soft_errors: Vec<RusticError>,
      hard_errors: Vec<RusticError>,
  }
  ```
- or return a tuple of `(T, Vec<RusticError>)` from a function, in case we want to handle multiple errors
- we could also go for making `RusticError` a trait and implement common behaviour on structs for each possible error struct

But this topic is definitely something we should talk about to find a good approach.

aawsome:
In principle I like the ides of `lazy_errors` - I however think that we want to collect not only errors, but also warnings - or, more general, a kind of ErrorLevel together with the error.
Also, we have two cases:
- situations where we want to collect errors, but continue, i.e. we want to return a result *and* a error list
- situation where we want to abort on errors, i.e. we want to return only the error list and indicate that things went wrong...

simonsan:
What about introducing an `RusticIssue` like that:

```rust
#[derive(Debug)]
pub enum RusticIssue {
    Error(RusticError),
    Warning {
        message: String,
    },
    Info {
        message: String,
    },
}
```

and then maybe something like:

```rust
#[derive(Debug)]
pub enum RusticResult<T> {
    Ok(T, Option<Vec<RusticIssue>>),  // For when we return a result along with possible issues
    Err(Vec<RusticIssue>),    // For when we abort and return only the issues that occured
}
```

simonsan:
I sketched out an idea, to fix the pain points you mentioned in a more generic way, than introducing special error handling/collection for `check` only. Could you take a look?

https://github.com/rustic-rs/rustic_core/pull/318

aawsome:
I think the main question here is:
Should we merge this PR with maybe just small modifications (quick&dirty) to fix the check issues and then overwork the error handling once we have a general error handling framework? Or should we wait for the framework?
I have a tendency towards a quick&dirty merge here, actually...

simonsan:
> I think the main question here is: Should we merge this PR with maybe just small modifications (quick&dirty) to fix the check issues and then overwork the error handling once we have a general error handling framework? Or should we wait for the framework? I have a tendency towards a quick&dirty merge here, actually...

I would wait, I don't think (no offence really, we are doing a lot here) this should be merged as it is. We are introducing new errors here, etc. We should just focus on this error handling topic, after the next `rustic` release and have it fixed long-term. This would just move it further away, while we feel we 'fixed' something, but we didn't really. We just moved a well-formed fix further away.

simonsan:
Let's rather do it right one time with a bit more effort now, than doing it more than once, just to be quick and dirty. 😅

simonsan:
> * situations where we want to collect errors, but continue, i.e. we want to return a result _and_ a error list

For this situation, can you give an example, because this is worth to explore why it is like that. There could be many reasons for it, e.g. architectural problems, prolonged error handling, instead of doing it locally, etc.

When it comes to collecting errors in a thread, we can go for a channel and handle it in the main thread.

aawsome:
> > * situations where we want to collect errors, but continue, i.e. we want to return a result _and_ a error list
> 
> For this situation, can you give an example, because this is worth to explore why it is like that. There could be many reasons for it, e.g. architectural problems, prolonged error handling, instead of doing it locally, etc.

An example is `Repository::get_snapshots`. This loads all Snapshotfiles from the repository and decrypts them into a `Vec<Snapshotfile>'. 

There are errors like not being able to list snapshots which will cause this command to fail. But each individual snapshot may fail either to load from backend or to decrypt (e.g. data corruption). Currently, this functions fails if any snapshot fails to load+decrypt. This means if we have a corrupt snapshot file, the repository can be basically no longer used until this is fixed.

We however may want to introduce resiliency: It is e.g. very nasty if the whole repository is not accessible just because that old and now unimportant snapshotfile has an issue.  If we introduce this, then the command would return the successful loaded snapshots and a list of failed snapshots with the error(s) - and we would at least need to indicate that something went wrong by not sending a 0 return code in some cases...

simonsan:
I would imagine it like this somehow (draft):

2024-10-14 22_00_49-snapshotfile rs (Working Tree) (snapshotfile rs) - rustic (Workspace) - Rust Dev

So we have a `Vec<RusticError>` that we return from `SnapshotFile::fill_missing`:

2024-10-14 22_04_30-snapshotfile rs (Working Tree) (snapshotfile rs) - rustic (Workspace) - Rust Dev

, and we reserve the `Err(E)` of `RusticResult` for hard errors.

But in the end, it's the question of why we want to return a list of the non-working snapshots and not just log an error there. What value has the information for the caller, that there are broken snapshots it cannot be operated on? We just don't want it to fail, right? If we return that `Vec<RusticError>` to the caller, what else will it do, than just logging it/printing to the user?
If we would want to handle the broken snapshots as well, we might rather return something like:
`RusticResult<Vec<RusticResult<SnapshotFile>>>` with outer Result being the operation itself that is fallible and indicates something went horribly wrong. And the inner Results being the Snapshots itself, that the caller wants to handle the Ok(T) values of and do something with the errors.

crates/core/src/commands/check.rs
aawsome: is changed.

crates/core/src/commands/check.rs
aawsome: fixed.

crates/core/src/commands/check.rs
aawsome: we currently have a `lock().unwrap()` for all `std::sync::Mutex` in the code. The solution is to use a implementation which is not affected by poisoning, but I think this should go into another PR.

aawsome:
@simonsan As having `check` report errors is actually an important step to get to a non-beta stage (for end-users but even more important for our integration tests), I tried to simplify this PR. 
It is now expicitely a `check`-only solution as having solved this is better than discussing global solutions where the discussion doesn't come to a result in near future..
I also added some integration tests (there are more to add, of course) and hope that these show why having a list of warnings/errors is desirable especially for the check result.
This is now a breaking change as `check` return now `CheckResults`.

PR #226, opened 2024-05-28
fix: Always sort StringList

aawsome:
We now use a sorted set for storing paths and tags used in snapshots. This ensures that independent from how they are saved, identical sets are always grouped together.

closes https://github.com/rustic-rs/rustic/issues/1158


simonsan (dismissed):

simonsan:
(Still some clippy fixes to be done, but otherwise looks good to me.)

Issue #227, opened 2024-06-08
Inability to fully specify type of `Repository`


TilBlechschmidt:
Hey there 👋 

I was trying to use `rustic_core` to read files from an on-disk backup and wanted to dynamically create and cache repositories (i.e. store them in a collection). However, while designing my internal API, I ran into the issue that the type of an **opened and fully indexed** Repository can not be fully specified:

```rust
use rustic_core::{
    repository::{FullIndex, IndexedStatus}, // <-- ERROR: These are inaccessible
    NoProgressBars, OpenStatus
};

type RusticRepository = rustic_core::Repository<NoProgressBars, IndexedStatus<FullIndex, OpenStatus>>;
```

The problem is that while the `repository::IndexedFull` trait is re-exported, the concrete types `repository::IndexedStatus` and `repository::FullIndex` are not.

When directly using the return value of `Repository::new(...).open().to_indexed()` this does not matter, since the methods can be called regardless. Returning this value from a function is also possible using `impl IndexedFull`. However, when for example constructing a wrapping struct like so:

```rust
pub struct RepositoryWrapper(Repository<NoProgressBars, _>);
```

It becomes impossible to define the type since the previously mentioned concrete types are not public. It is possible to work around this by using generics and the `IndexedFull` trait.

However, this forces one to push the creation of the repository far up the call stack (i.e. passing it in as a parameter in the code below) and litters the entire codebase with generics that will only ever have one representation. It also becomes impossible to dynamically create and store Repositories in a collection.

```rust
pub struct RepositoryWrapper<S>(Repository<NoProgressBars, S>);

impl<S> RepositoryWrapper<S> {
    pub fn new() -> Self<S> {
        // ERROR: S does not match the concrete but opaque type returned by this since `S` needs to be specified by the callee!
        Self(Repository::new(...).open().to_indexed())
    }
}
```

I also tried using dynamic dispatch instead by wrapping the repository in a `Box` or `Arc` and using `dyn IndexedFull`. However, this runs into the same issue where the associated type parameter `I` needs to be specified which resolves to `index::GlobalIndex` which unfortunately is also private.

---

My suggestion would be to re-export the two types `FullIndex` and `IndexedStatus` in [`lib.rs#L151`](https://github.com/rustic-rs/rustic_core/blob/f1f73dbfa3b47f65fb55da2ba45062de32436bc1/crates/core/src/lib.rs#L151) as follows:

```diff
pub use crate::{
    ...
[...]

PR #229, opened 2024-06-14
fix: Export types so the Repository type can be fully specified

TilBlechschmidt:
See #227 for more details.

I've added a test-case which constructs a minimal wrapper type around a Repository that implements `IndexedFull`, puts two instances into a collection (the primary motivator behind this), and hands them to a function that relies on the trait.


TilBlechschmidt:
Updated to fix clippy warning, may or may not have forgotten to run that 🙈

aawsome (approved):
LGTM

Issue #232, opened 2024-07-28
Unable to use sshfs (FUSE) to backup files with rustic.


MarlonJMejia:
# Version of rustic

```bash
rustic --version
rustic 0.7.0
```

# What I am trying to accomplish

I am trying to mount a directory via sshfs to backup from a remote source.
The mount is successful and I am able to view the files for the remote host.  I've also tried to do a more bare bones mount with no options. (There is no error on the output or success after an hour)

user is `root` for both local and remote host to execute the commands and mountfiles.

```
sshfs -o 'reconnect,cache=no,noauto_cache,entry_timeout=0' user@remotehost:/directory /localdirectory/.backup_sshfs/

```

# The issue

I receive multiple warnings (see below), I initially though this warnings were simply warnings but it seems that backups are not sucessfull if there is a huge amount of files (or there is an actual error with sshfs) in the directory of the remote source.

## errors
```
[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`
[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`
```

## wc of files in the remote directorys to backup

```
fdfind -u . /mnt/storage/appdata/ -t f -t d | wc 
  49496   49501 4091785

fdfind -u . /var/lib/vz/dump/ -t f -t d | wc
    15      15     929
```

## Command to execute backup

```
rustic -r . backup /tmp/.backup_sshfs/mnt/storage/appdata /tmp/.backup_sshfs/var/lib/vz/dump/
```

```bash 
[00:00:00] backing up...                  ████████████████████████████████████████        0 B            0 B/s                            [WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`

[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`

[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`

[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`

[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`

[WARN] ignoring error `Os { code: 95, kind: Uncategorized, message: "Operation not supported" }`
```

# Notes

This same process works fine with Kopia, Duplicacy and Borg.

Backup process seems to work completely fine with a lower of total amount of files.

[...]

PR #233, opened 2024-07-28
fix: backup file even if failed listing extended attributes

nardoor:
This should fix rustic-rs/rustic_core#232 
However I'm not so sure as the error message I encountered is slightly different from what reported by the issue creator.
In both cases I think this PR is worth looking into when you have time.

## The idea:

Even on a system that supports extended-attributes, we might encounter some part of the file-system that doesn't.
For instance a `mount` with the right option (streams_interface=windows for instace) will not support xattr.
AFAIK it also seems to be the case with a simple sshfs on a debian/ubuntu (as experienced in  #232).

## Note 

I used a debug log to print the XAttr error in order to avoid erroring for every file of an extended attribute incompatible backup.

## Testing

I am once again unsure on what tests to add. 
Please let me know if anything comes up to your mind.

### DIFF
--- a/crates/core/src/backend/ignore.rs
+++ b/crates/core/src/backend/ignore.rs
@@ -16,7 +16,7 @@ use chrono::TimeZone;
 use chrono::{DateTime, Local, Utc};
 use derive_setters::Setters;
 use ignore::{overrides::OverrideBuilder, DirEntry, Walk, WalkBuilder};
-use log::warn;
+use log::{debug, warn};
 #[cfg(not(windows))]
 use nix::unistd::{Gid, Group, Uid, User};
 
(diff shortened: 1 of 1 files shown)


crates/core/src/backend/ignore.rs
```diff
@@ -531,29 +566,19 @@ fn map_entry(
...
+    let extended_attributes = match list_extended_attributes(entry.path()) {
+        Err(e) => {
+            // TODO - discuss this log:
+            // Most probably that if Xattr read failure happens for one file,
+            // it will also happen for every file in the same subdirectory.
+            //
+            // The end result would be a real flood of errors.
+            // For this reason, I propose to have this log at "debug" level
+            // so it's not transparent, but doesn't impact a "lambda" use case of rustic.
+            debug!("ignoring error: {e}\n");
```
aawsome, line 578: So far we throw a warning if there is an error with a specific file - even though this error could come multiple times. So I would also use `warn` here. In future, we might want to collect errors and give some consolidated result at the end, but ATM I think a warning is fine.

aawsome (changes requested):
Thanks @nardoor for this PR! 
I think we should only change the loglevel (and remove the comment about the log-discussion)

nardoor:
Hello @aawsome ,

Thanks for taking the time to review. I'll fix the cross compilation issue and apply the requested change soon. 

I still personally think a warning level for this log might harm some user's UX. 
But I understand that the policy is to warn about recovered errors. 
Also other users might wonder why their `xattr` are not backed up.

crates/core/src/backend/ignore.rs
```diff
@@ -472,6 +472,41 @@ fn get_group_by_gid(gid: u32) -> Option<String> {
     }
 }
 
+#[cfg(target_os = "openbsd")]
```
aawsome, line 475: I think `#[cfg(all(not(windows),target_os = "openbsd"))]`should be used here

crates/core/src/backend/ignore.rs
```diff
@@ -472,6 +472,41 @@ fn get_group_by_gid(gid: u32) -> Option<String> {
...
+/// List [`ExtendedAttribute`] for a [`Node`] located at `path`
+///
+/// # Argument
+///
+/// * `path` to the [`Node`] for which to list attributes
+///
+/// # Errors
+///
+/// * [`IgnoreErrorKind::ErrorXattr`] - if Xattr couldn't be listed or couldn't be read
+#[cfg(not(target_os = "openbsd"))]
```
aawsome, line 489: same here about `not(windows)`

crates/core/src/backend/ignore.rs
nardoor: yeah, I forgot to backport the (caller) `map_entry` `cfg`  to the functions I extracted.

This explains the cross compilation error.
[66c1aed](https://github.com/rustic-rs/rustic_core/pull/233/commits/66c1aed810214b27924c8a69477dbbcd1472b0c1) should solve this

crates/core/src/backend/ignore.rs
nardoor: same: 

> yeah, I forgot to backport the (caller) `map_entry` `cfg`  to the functions I extracted.
> 
> This explains the cross compilation error.
> [66c1aed](https://github.com/rustic-rs/rustic_core/pull/233/commits/66c1aed810214b27924c8a69477dbbcd1472b0c1) should solve this

crates/core/src/backend/ignore.rs
nardoor: agreed and applied in [ab9f4ac](https://github.com/rustic-rs/rustic_core/pull/233/commits/ab9f4acdcaf507c9c4f37b59c2a39160433ba252)

nardoor:
Push forced the branch after I rebased on upstream/main

aawsome (approved):

PR #240, opened 2024-08-11
fix: parse commands given by arg or env using shell_words

aawsome:
closes https://github.com/rustic-rs/rustic/issues/1177


nardoor:
https://github.com/nardoor/rustic_core/commit/84f70d619c3ac5b25ae4f2ee54fb2f0b724b0da4 
These small changes might help 😊

PR #247, opened 2024-08-26
fix: don't give invalid password error for other keyfile errors

aawsome:


crates/core/src/repofile/keyfile.rs
```diff
@@ -333,8 +334,12 @@ pub(crate) fn find_key_in_backend<B: ReadBackend>(
         key_from_backend(be, id, passwd)
     } else {
         for id in be.list(FileType::Key).map_err(RusticErrorKind::Backend)? {
-            if let Ok(key) = key_from_backend(be, &id, passwd) {
-                return Ok(key);
+            match key_from_backend(be, &id, passwd) {
+                Ok(key) => return Ok(key),
+                Err(RusticError(RusticErrorKind::Crypto(
+                    CryptoErrorKind::DataDecryptionFailed(_),
```
simonsan, line 340: Not sure if it wouldn't be better to implement an accessor for this that exposes the underlying error for the match and use a match guard for it? So we don't use pub crate here?

aawsome: I think that a `pub crate` will allow refactoring without breaking any public API, so IMO it is fine.
Actually to do unit tests, we would (for best coverage of error) basically need all error variants accessible by test functions. I don't know if using accessors for all that is a good idea - therefore I favored the `pub crate`.

simonsan (approved):

PR #256, opened 2024-09-05
refactor!: Use different Id types

aawsome:
Breaking change:
Use different `Id` types for all places where `Id`s play a role. This makes the source code much more understandable.

This also prevents some nasty kind of bugs. This was possible before:
```rust
let snap = repo.get_snapshot_from_str("latest", |_| true)?;
let node = repo.node_from_path(snap.id, "/my/path".into())?;
```
But this wasn't working as `snap.id` is a snapshot id and `node_from_path` requires a tree id. This code will now fail at compile time.


aawsome:
@simonsan I think this is ready for a review. I did not finish changing all internals where `PackedId` can be used, but the interface should now just use the Id newtypes.

crates/core/src/backend/decrypt.rs
```diff
@@ -31,6 +31,8 @@ pub trait DecryptFullBackend: DecryptWriteBackend + DecryptReadBackend {}
 
 impl<T: DecryptWriteBackend + DecryptReadBackend> DecryptFullBackend for T {}
 
+pub type StreamResult<Id, F> = RusticResult<Receiver<RusticResult<(Id, F)>>>;
```
simonsan, line 34: Do we need to export this for public API?

crates/core/src/blob.rs
```diff
@@ -66,17 +66,46 @@ impl<T: Default> Initialize<T> for BlobTypeMap<T> {
     }
 }
 
+new_id!(BlobId, "blob");
+
+pub trait PackedId: Copy + Into<BlobId> + From<BlobId> {
```
simonsan, line 71: Also here, does it need to be part of the public API?

crates/core/src/blob.rs
```diff
@@ -66,17 +66,46 @@ impl<T: Default> Initialize<T> for BlobTypeMap<T> {
...
 /// A `Blob` is a file that is stored in the backend.
 ///
 /// It can be a `tree` or a `data` blob.
 ///
 /// A `tree` blob is a file that contains a list of other blobs.
 /// A `data` blob is a file that contains the actual data.
-#[derive(Debug, PartialEq, Eq, Clone, Constructor)]
-pub(crate) struct Blob {
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Constructor)]
+pub struct Blob {
```
simonsan, line 105: Part of public API?

crates/core/src/blob/packer.rs
```diff
@@ -20,16 +20,17 @@ use crate::{
...
         configfile::ConfigFile,
         indexfile::{IndexBlob, IndexPack},
-        packfile::{PackHeaderLength, PackHeaderRef},
+        packfile::{PackHeaderLength, PackHeaderRef, PackId},
         snapshotfile::SnapshotSummary,
     },
 };
 
+use super::BlobId;
+
```
simonsan, line 33: Relative import

crates/core/src/blob.rs
```diff
@@ -66,17 +66,46 @@ impl<T: Default> Initialize<T> for BlobTypeMap<T> {
...
 
+new_id!(BlobId, "blob");
+
+pub trait PackedId: Copy + Into<BlobId> + From<BlobId> {
+    const TYPE: BlobType;
+}
+
+#[macro_export]
+/// Generate newtypes for `Id`s identifying packed blobs
+macro_rules! new_blobid {
```
simonsan, line 77: Would name it something like `impl_blob_id` to make it easier readable and make people aware, that this just adds `impl` to a type.

crates/core/src/id.rs
```diff
@@ -16,6 +16,33 @@ pub(super) mod constants {
     pub(super) const HEX_LEN: usize = LEN * 2;
 }
 
+#[macro_export]
+/// Generate newtypes for `Id`s identifying Repository files
+macro_rules! new_id {
```
simonsan, line 21: Would name it `define_new_id_struct`, so it's clear, here is a new struct being defined.

crates/core/src/repofile.rs
```diff
@@ -10,9 +12,41 @@ pub(crate) mod snapshotfile;
...
+
+/// Marker trait for Ids which identify repository files
+pub trait RepoId: Deref<Target = Id> + From<Id> + Sized + Send + Sync + 'static {
+    /// The [`FileType`] associated with Id type
+    const TYPE: FileType;
+}
+
+#[macro_export]
+/// Generate newtypes for `Id`s identifying Repository files
+macro_rules! new_repoid {
```
simonsan, line 27: Would name it `impl_repo_id` to make us aware, that it just add an implementation for RepoId

crates/core/src/repofile.rs
```diff
@@ -10,9 +12,41 @@ pub(crate) mod snapshotfile;
...
+    ($a:ident, $b: expr) => {
+        $crate::new_id!($a, concat!("repository file of type", stringify!($b)));
+        impl $crate::repofile::RepoId for $a {
+            const TYPE: FileType = $b;
+        }
+    };
+}
+
+#[macro_export]
```
simonsan, line 36: Would name it `impl_repo_file` to make us aware, that it just adds an implementation for RepoFile

simonsan (changes requested):
Great job! That looks so much cleaner now! A few nits:

crates/core/src/blob.rs
aawsome: Yes, it is used in `Repository::get_index_item`.

crates/core/src/backend/decrypt.rs
aawsome: no, is changed.

crates/core/src/blob.rs
aawsome: no, is changed.

crates/core/src/blob/packer.rs
aawsome: is fixed.

crates/core/src/blob.rs
aawsome: renamed.

crates/core/src/id.rs
aawsome: renamed.

crates/core/src/repofile.rs
aawsome: renamed.

crates/core/src/repofile.rs
aawsome: renamed.

crates/core/src/blob.rs
simonsan: Do we then need to export it directly in `lib.rs`?

simonsan (approved):

PR #257, opened 2024-09-05
feat: Add autocompletion hints 

nardoor:
## Feature 

**Add autocompletion value hints**

This helps some shells to propose better completions.
For instance on zsh, no completion were proposed when doing rustic backup <...>[TAB]

## Hints added

- `rustic --repository <...>[TAB]`, `rustic -r <...>[TAB]`, `rustic --repo <...>[TAB]`
- `rustic backup --as-path [TAB] <...>`
- `rustic backup --stdin-filename [TAB] <...>` 
- `rustic backup --description-from <...>[TAB]`
- `rustic --password-file <...>[TAB]`
- `rustic --cache-dir <...>[TAB]`


nardoor:
similar / complementary changes in rustic-rs/rustic#1225

nardoor:
- [ ] The failed `lint` is out of this PR's code, but I can fix it anyways.

- [ ] For the failed tests, this evening I'll have a look at why my code impacted those snapshots and maybe re-generate them if that makes sense.

aawsome:
About the clippy and test failures, I'll make a separate PR.

aawsome:
> * For the failed tests, this evening I'll have a look at why my code impacted those snapshots and maybe re-generate them if that makes sense.

I think this is due to the fact that we have snapshots with exactly the identical timestamp in those tests. The algorithm uses `sort_unstable` which is kind of non-predictable for this input. I think best we simply adapt the test cases to have unique timestamps.

PR #262, opened 2024-09-08
feat: Add read-data-subset to CheckOptions; allow to check given trees

aawsome:
Adds `Repository::check_with_trees` which allows to do tree and data check only for the given trees.

Moreover, `CheckOptions` now have the `read_data_subset` option which allows to read only parts of the pack files.

Also warming up of packfiles in case of hot/cold respository has been added.

closes https://github.com/rustic-rs/rustic/issues/251


simonsan (dismissed):

crates/core/src/id.rs
```diff
@@ -169,6 +175,11 @@ impl Id {
         let mut vec = vec![0; length];
         r.read_exact(&mut vec).is_ok() && self == &hash(&vec)
     }
+
+    /// returns the first 4 bytes as u32 (interpreted as little endian)
+    pub fn as_u32(&self) -> u32 {
```
simonsan, line 180:
```suggestion
    #[must_use]
    pub fn as_u32(&self) -> u32 {
```

simonsan (approved):

PR #263, opened 2024-09-08
fix: properly finish progress bar in Repository::get_snapshot_group

aawsome:
Also adds a few more integration tests for snapshot gathering methods.


aawsome:
@simonsan Yes. Unfortunately I don't have a windows machine to run them. And the CI only outputs the first missing/changed snapshots; in case there are multiple snapshots, I have to run the CI multiple times which is actually quite annoying.. Do you know how this can be improved?

simonsan:
> @simonsan Yes. Unfortunately I don't have a windows machine to run them. And the CI only outputs the first missing/changed snapshots; in case there are multiple snapshots, I have to run the CI multiple times which is actually quite annoying.. Do you know how this can be improved?

I imagine the reason for this may be, that we stop on first error (not quite sure though, this is the case), so we would need to run all tests and not stop on failed tests (I guess?). At least worth a try.

simonsan (approved):

PR #269, opened 2024-09-19
feat!: Use CommandInput for commands

aawsome:
This re-allows to use a String for `password-command` or `warm-up-command`.

Breaking change: The former version to give the command in an array is no longer supported.
Instead of
```
password-command = ["echo", "password"]
```
please use either of
```
password-command = "echo password"
password-command = { command = "echo", args = ["password"] }
```


simonsan:
Added to https://github.com/rustic-rs/docs/pull/88

crates/backend/src/rclone.rs
```diff
@@ -150,12 +148,16 @@ impl RcloneBackend {
         let user = Alphanumeric.sample_string(&mut thread_rng(), 12);
         let password = Alphanumeric.sample_string(&mut thread_rng(), 12);
 
-        let mut rclone_command =
-            split(rclone_command.map_or("rclone serve restic --addr localhost:0", String::as_str))?;
-        rclone_command.push(url.as_ref().to_string());
+        let mut rclone_command = rclone_command.map_or(
+            "rclone serve restic --addr localhost:0".to_string(),
```
simonsan, line 152: Would it make sense, to take this hardcoded command out as a constant `RCLONE_COMMAND` to `mod constants {}` in `rclone.rs` so we have this easily accessible and separated from the rest of the code?

crates/core/src/repository.rs
```diff
@@ -356,11 +341,11 @@ impl<P> Repository<P, ()> {
         let mut be = backends.repository();
         let be_hot = backends.repo_hot();
 
-        if !opts.warm_up_command.is_empty() {
-            if opts.warm_up_command.iter().all(|c| !c.contains("%id")) {
+        if let Some(warm_up) = &opts.warm_up_command {
+            if !warm_up.args().iter().all(|c| !c.contains("%id")) {
```
simonsan, line 345: Why is this negated now? 🤔

crates/backend/src/rclone.rs
aawsome: is changed.

crates/core/src/repository.rs
aawsome: Good catch - that was a typo.
I fixed it and also added an integration test which would have caught this.

simonsan (approved):

PR #267, opened 2024-09-16
fix!: use plural names for options

aawsome:
Consistently uses singular and plural names for options in config and CLI

BREAKING CHANGE: This requires to adapt CLI calls and config keys.

Affected options:
- `glob` -> `globs` in config profile
- `iglob` -> `iglobs` in config profile
- `glob`-file -> `glob-files` in config profile
- `iglob-file` -> `iglob-files` in config profile
- `custom-ignore-file` -> `custom-ignore-files` in config profile
- `tag`-> `tags` in config profile
- `keep-tags` -> now only array
- `keep-ids` -> now only array

PR #270, opened 2024-09-22
fix: de/serialize tags as DisplayFromStr

aawsome:
Fixes a regression about de/serializing tags, introduced by #267

### DIFF
--- a/crates/core/src/repofile/snapshotfile.rs
+++ b/crates/core/src/repofile/snapshotfile.rs
@@ -53,6 +53,7 @@ pub struct SnapshotOptions {
     pub label: Option<String>,
 
     /// Tags to add to snapshot (can be specified multiple times)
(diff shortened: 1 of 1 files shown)


nardoor (commented):
1. I acknowledge this was a breaking change. Despite that, could there be some "test" that would have caught this "regression" ? 

2. A quick and naive look into #267 shows many similar `#[serd_as(...)]` suppression. I assume some might be intended. Could you make sure no similar regression were brought to other fields ?

aawsome:
About 1. The tests which did catch this regression (even though breaking, I changed them such that the desired config profiles should run) are located in rustic.

About 2. The tests did not find other things. The removal of `OneOrMany` was intentional, the problem for tags was the removal of `OneOrMany<DisplayFromStr>` which also removed the `DisplayFromStr` de/serialization...

simonsan (approved):

PR #285, opened 2024-09-29
fix: Don't query the default cache directory when a custom one is set

chenxiaolong:
The default cache directory is now only queried if no explicit cache path is given. This is necessary to prevent errors on platforms where the default cache directory does not exist, like Android.

### DIFF
--- a/crates/core/src/backend/cache.rs
+++ b/crates/core/src/backend/cache.rs
@@ -237,11 +237,13 @@ impl Cache {
     /// [`CacheBackendErrorKind::NoCacheDirectory`]: crate::error::CacheBackendErrorKind::NoCacheDirectory
     /// [`CacheBackendErrorKind::FromIoError`]: crate::error::CacheBackendErrorKind::FromIoError
     pub fn new(id: RepositoryId, path: Option<PathBuf>) -> RusticResult<Self> {
-        let mut path = path.unwrap_or({
+        let mut path = if let Some(p) = path {
+            p
+        } else {
             let mut dir = cache_dir().ok_or_else(|| CacheBackendErrorKind::NoCacheDirectory)?;
             dir.push("rustic");
             dir
-        });
+        };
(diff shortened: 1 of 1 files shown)


aawsome (approved):
LGTM

aawsome:
@chenxiaolong Good catch, thanks a lot for the PR!

simonsan:
@chenxiaolong Good one! Can you apply the fixes `clippy` proposed ( https://github.com/rustic-rs/rustic_core/actions/runs/11096543748/job/30852746926?pr=285#step:5:123 ) and push the changes please? 🚀

crates/core/src/backend/cache.rs
```diff
@@ -237,11 +237,14 @@ impl Cache {
...
-            dir
-        });
+        let mut path = match path {
+            Some(p) => p,
+            None => {
+                let mut dir = cache_dir().ok_or_else(|| CacheBackendErrorKind::NoCacheDirectory)?;
+                dir.push("rustic");
+                dir
+            }
+        };
```
simonsan, line 247: Clippy suggest due to `single match`:

```rust
let mut path = if let Some(p) = path { p } else {
    let mut dir = cache_dir().ok_or_else(|| CacheBackendErrorKind::NoCacheDirectory)?;
    dir.push("rustic");
    dir
};
```

simonsan (changes requested):

crates/core/src/backend/cache.rs
chenxiaolong: Sure thing, fixed!

chenxiaolong:
Ah crap, `cargo clippy --fix` used formatting different from `cargo fmt`. Fixing

simonsan (approved):

simonsan:
> Ah crap, `cargo clippy --fix` used formatting different from `cargo fmt`. Fixing
> 
> EDIT: Fixed!

Thank you for the contribution! 🚀

PR #291, opened 2024-10-01
fix(backend)!: Use correct merge stratgy for repository options

simonsan:
This fixes the merge strategy for values of `options`, `options_hot`, and `options_cold` to ignore incoming elements if they are already existing in the HashMap. The prior implementation was overwriting the values.

Breaking change: The freestanding and public `extend` function has been removed. You can use `conflate::hashmap::overwrite` as a replacement.


aawsome (approved):
LGTM

PR #295, opened 2024-10-02
ci: don't run heavy ci on push to main

simonsan:
Due to extensive use of the merge queue, we essentially run the heavy ci twice. once in the merge group and then again when being pushed from the merge group to main. This wastes runner resources and time. Instead we can just run on schedule every night and count on the merge group CI runs.

### DIFF
--- a/.github/workflows/ci-heavy.yml
+++ b/.github/workflows/ci-heavy.yml
@@ -15,16 +15,12 @@ defaults:
     shell: bash
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


simonsan:
We should also change it, because we only push to `main` when it's an important fix or some CI related things. Pushing to `main` also include unsetting `Do not allow bypassing the above settings` in the branch protection settings, which is another hurdle. So I think this is fine.

PR #297, opened 2024-10-02
fix(errors): better error message for hot/cold repo in check

simonsan:
Fixes https://github.com/rustic-rs/rustic/issues/1229

### DIFF
--- a/crates/core/src/commands/check.rs
+++ b/crates/core/src/commands/check.rs
@@ -419,7 +419,7 @@ fn check_packs(
 
     if let Some(hot_be) = hot_be {
         let p = pb.progress_spinner("listing packs in hot repo...");
-        check_packs_list(hot_be, tree_packs)?;
+        check_packs_list_hot(hot_be, tree_packs, &packs)?;
         p.finish();
     }
 
@@ -431,7 +431,7 @@ fn check_packs(
 }
 
 // TODO: Add documentation
-/// Checks if all packs in the backend are also in the index
+///Checks if all packs in the backend are also in the index
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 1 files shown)


crates/core/src/commands/check.rs
```diff
@@ -447,7 +447,9 @@ fn check_packs_list(be: &impl ReadBackend, mut packs: HashMap<PackId, u32>) -> R
         .map_err(RusticErrorKind::Backend)?
     {
         match packs.remove(&PackId::from(id)) {
-            None => warn!("pack {id} not referenced in index. Can be a parallel backup job. To repair: 'rustic repair index'."),
+            None => {
+                warn!("(hot) pack {id} not referenced as tree pack in index, but as data pack.");
```
aawsome, line 451: this is currently used for both hot packs and for normal (maybe cold) packs. So changing this this way makes it correct for hot packs, but wrong in the standard case.

simonsan: Ah! How would we differentiate in `None` between hot and cold packs, then I can implement it.

aawsome: I copied the check_pack function and now handle checking of hot packs differently from normal packs. Also added a test about data packs contained in the hot repo.

simonsan: Is duplicating the function and code the only good way to handle it, currently? Do you think we will implement different handling (checking) for hot and cold files in the future, so we could reuse this?

aawsome: It is already different. In the hot pack check we now additionally check against the total pack list to give another error in the specific situation the user had...

crates/core/src/commands/check.rs
```diff
@@ -456,6 +456,46 @@ fn check_packs_list(be: &impl ReadBackend, mut packs: HashMap<PackId, u32>) -> R
     Ok(())
 }
 
+///Checks if all packs in the backend are also in the index
```
simonsan, line 459:
```suggestion
/// Checks if all packs in the backend are also in the index
```

crates/core/src/commands/check.rs
```diff
@@ -426,7 +426,7 @@ fn check_packs(
 }
 
 // TODO: Add documentation
-/// Checks if all packs in the backend are also in the index
+///Checks if all packs in the backend are also in the index
```
simonsan, line 429:
```suggestion
/// Checks if all packs in the backend are also in the index
```

PR #300, opened 2024-10-03
refactor(deps)!: upgrade to new conflate version

simonsan:
BREAKING CHANGE: We replaced the use of HashMaps with BTreeMaps in `rustic_backend`.


aawsome (approved):
LGTM.

I just think we should wait until we are ready for a new breaking release ;-)

PR #305, opened 2024-10-04
test(commands): command input parsing

simonsan:
Signed-off-by: simonsan <14062932+simonsan@users.noreply.github.com>

### DIFF
--- a/crates/core/src/repository/command_input.rs
+++ b/crates/core/src/repository/command_input.rs
@@ -224,3 +224,24 @@ impl OnFailure {
 fn split(s: &str) -> RusticResult<Vec<String>> {
     Ok(shell_words::split(s).map_err(|err| RusticErrorKind::Command(err.into()))?)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    #[rstest]
+    #[case("echo hello", "echo", &["hello"])]
+    #[case("echo hello >> /tmp/hello", "echo", &["hello", ">>", "/tmp/hello"])]
+    fn test_command_input_passes(
+        #[case] input: &str,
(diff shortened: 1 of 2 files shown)


crates/core/src/repository/command_input.rs
```diff
@@ -220,7 +221,42 @@ impl OnFailure {
...
     Ok(shell_words::split(s).map_err(|err| RusticErrorKind::Command(err.into()))?)
 }
+
+/// helper to split arguments
+///
+// We keep the return type as `RusticResult<Vec<String>>` to keep the internal api
+// consistent with the other platforms.
+#[allow(clippy::unnecessary_wraps)]
+#[cfg(windows)]
+fn split(s: &str) -> RusticResult<Vec<String>> {
```
aawsome, line 235: with this, this is a change and not only a test PR ;-)

I experimented shortly on winsplit. One drawback I found was that it seems to be impossible to have the reverse operation of `split` - this makes a suitable `Display` impl quite hard.

Also I just wanted to note that a OS-dependent parsing may cause irritation for users using both windows and non-windows. I don't use windows, but are there cases where `shellwords::split` don't work on windows commands?

aawsome:
just one thing: Did you see the tests in `crates/core/tests/command_input.rs`?

simonsan:
> just one thing: Did you see the tests in `crates/core/tests/command_input.rs`?

Thanks for the heads-up, no I didn't, I was actually wondering!

crates/core/src/repository/command_input.rs
simonsan: Yeah, I'm playing around with it to get a feeling for the parsing. It's true, that we need to be attentive here.

PR #307, opened 2024-10-05
Lock more

aawsome:
see https://github.com/rustic-rs/rustic/discussions/1050

BREAKING CHANGES:
-    DeleteOptions in SnapshotFile now has more enum options.
-    Repository::save_snapshots now returns the list of Ids of the saved snapshots.

Missing points:
- [ ] Handle duplicate snapshots correctly - especially create at max. 1 new snapshot per "original" snapshot during lock
- [ ] Handle locks in forget command (retention policy: only count one/the original snapshot for retention)
- [ ] Handle locks correctly in the tag command (don't even try to remove locked snapshots)


simonsan:
Blocked on #163: PR requires some changes and maybe a discussion about `Lock` design as a type state for a backend? Not entirely sure, if that would be a good approach as the expectation on a backend with a lock is not completely clear to me.

PR #314, opened 2024-10-05
refactor(errors): improve message and add logging when sending tree from backend panics

### LINKED ISSUE #239: panic on prune
Hi,
thanks for #235 , got a new crash now:

```
name = "rustic-rs"
operating_system = "Arch Linux [64-bit]"
crate_version = "0.7.0-dev"
explanation = """
Panic occurred in file '/data/build/.cargo/git/checkouts/rustic_core-fa42f42e094d79df/1cecce8/crates/core/src/blob/tree.rs' at line 690
"""
cause = 'called `Result::unwrap()` on an `Err` value: "SendError(..)"'
method = "Panic"
backtrace = """

   0: 0x61ebd4d5e1da - <unresolved>
   1: 0x61ebd4d5b7c7 - <unresolved>
   2: 0x61ebd4ffa9cb - <unresolved>
   3: 0x7335ac51839d - <unresolved>
   4: 0x7335ac59d49c - <unresolved>
   5:        0x0 - <unresolved>"""
```

simonsan:
Related #239
Related #209

### DIFF
--- a/crates/core/src/blob/tree.rs
+++ b/crates/core/src/blob/tree.rs
@@ -13,6 +13,7 @@ use derive_setters::Setters;
 use ignore::overrides::{Override, OverrideBuilder};
 use ignore::Match;
 
+use log::debug;
 use serde::{Deserialize, Deserializer};
 use serde_derive::Serialize;
 
@@ -689,9 +690,14 @@ impl<P: Progress> TreeStreamerOnce<P> {
             let out_tx = out_tx.clone();
             let _join_handle = std::thread::spawn(move || {
                 for (path, id, count) in in_rx {
-                    if let Err(err) = out_tx
-                        .try_send(Tree::from_backend(&be, &index, id).map(|tree| (path, tree, count)))
-                    {
+                    debug!(
+                        "Loading tree {id} from {} with count {count}",
+                        path.display()
+                    );
+
+                    if let Err(err) = out_tx.try_send(
+                        Tree::from_backend(&be, &index, id).map(|tree| (path, tree, count)),
+                    ) {
                         panic!("
                         Sending tree {id} on channel failed: {err}. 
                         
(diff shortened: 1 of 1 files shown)


crates/core/src/blob/tree.rs
```diff
@@ -689,9 +690,11 @@ impl<P: Progress> TreeStreamerOnce<P> {
             let out_tx = out_tx.clone();
             let _join_handle = std::thread::spawn(move || {
                 for (path, id, count) in in_rx {
-                    out_tx
+                    if let Err(err) = out_tx
                         .send(Tree::from_backend(&be, &index, id).map(|tree| (path, tree, count)))
-                        .unwrap();
+                    {
+                        error!("Error sending tree {id} from backend: {err}");
```
aawsome, line 696: This makes rustic not panic anymore. But it will return `OK(())` from this method. I wonder if this might lead to rustic not performing a requested operation while only showing this error log but not returing any error...
Is there a way that we can also return `Err` here?

aawsome: Might by actually something to discuss in our general error topic: Handle errors in threads we just have spawned...

simonsan: I think we should use a channel to communicate with the main thread and send the errors there. In General. In this case, it's harder to judge. I think we need to panic!() here in one way, as I think a `send` not working out might be a bug in the program. So I agree, this might not be a good solution to just log.

simonsan: Panicking, but improved the message for the user by using `try_send`. We should investigate this more, this shouldn't panic if the channel `is connected` and `not full`. So either way, I improved this in a way, we are able to further diagnose the error in the future, should it happen again.

aawsome (approved):

PR #316, opened 2024-10-05
feat: replace log with tracing and use tracing::instrument where useful

### LINKED ISSUE #315: use `tracing` crate for logging
I think it would be nice to use the `tracing` crate throughout `rustic_core`, as we could make use of it also regarding the `#[instrument]` macro. Giving us an easy opportunity to track calls into methods/functions as seen here: https://github.com/tokio-rs/tracing?tab=readme-ov-file#in-libraries

> the `#[tracing::instrument]` attribute creates and enters a span
> every time the instrumented function is called. The span is named after
> the function or method. Parameters passed to the function are recorded as fields.

This will make debugging much easier for downstream users e.g. in `rustic`, and we will also get much more information and thus insight into (future?) asynchronous code.

The nice thing about the migration: it uses the same `log` facade, so we can just start annotating methods with `tracing::instrument`.

simonsan:
Fixes #315 

TODO:
- [x] replace log with tracing
- [ ] annotate functions/methods with `tracing::instrument`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,15 @@ rust-version = "1.76.0"
 aho-corasick = "1.1.3"
 anyhow = "1.0.89"
 bytes = "1.7.2"
+derive_setters = "0.1.6"
+displaydoc = "0.2.5"
 enum-map = "2.7.3"
 rustic_backend = { path = "crates/backend" }
 rustic_core = { path = "crates/core", version = "0" }
 rustic_testing = { path = "crates/testing", version = "0" }
 simplelog = "0.12.2"
+thiserror = "1.0.64"
+tracing = "0.1.40"
 
 # dev-dependencies
 rstest = "0.23.0"
--- a/crates/backend/Cargo.toml
+++ b/crates/backend/Cargo.toml
@@ -44,16 +44,16 @@ rclone = ["rest", "dep:rand", "dep:semver"]
 rustic_core = { workspace = true }
 
 # errors
-anyhow = "1.0.89"
-displaydoc = "0.2.5"
-thiserror = "1.0.64"
+anyhow = { workspace = true }
+displaydoc = { workspace = true }
+thiserror = { workspace = true }
 
 # logging
-log = "0.4.22"
+tracing = { workspace = true }
 
 # other dependencies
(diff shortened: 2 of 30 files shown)


aawsome:
While I agree that we might want to migrate from `log` to `tracing`, I actually would do so when we have decided for the architecture we are using to run rustic in some optimal `sync/async/parallelized` way. Then, it will be a first implementation step to go into this direction.

simonsan:
Hmm, I don't understand, why would that be related to an architectural decision? `tracing` is essential in `async/parallelized` contexts, for sure. But we can already profit a lot from it by just instrumenting some essential functions/methods so we get more context in our logging. Do you see any downside there? 🤔

aawsome:
> Hmm, I don't understand, why would that be related to an architectural decision? `tracing` is essential in `async/parallelized` contexts, for sure. But we can already profit a lot from it by just instrumenting some essential functions/methods so we get more context in our logging. Do you see any downside there? 🤔

No downside. I just don't see any benefit we currently can get. That might change and `tracing` may help us finding bugs during a big refactor. But currently I just don't see the point in doing so. Better focus on topics like error handling..

simonsan:
(Sorry, I accidentally edited your comment instead of answering! 😅)

I collected all the advantages in here: https://github.com/rustic-rs/rustic_core/issues/315

The benefits are structured logging, the `tracing::instrument` macro, and others. I think it's not a big effort, to be honest. It's just replacing `log::` with tracing (log facade) and start to instrument certain methods and functions where we are often asking us what the input values to that functions were. For example, tree archiver. Everything where we already parallelize, for example to find race conditions/send errors:

- https://github.com/rustic-rs/rustic_core/pull/314
- https://github.com/rustic-rs/rustic_core/pull/306

simonsan:
I started out to instrument the functions, that are related to:

- async, e.g. `WebDavFS` and `opendal` backend
- in general multithreading (`TreeArchiver`)
- contain `*_par_*`/`readahead*` via `rayon`
- send/recv data on `channels`


Follow up: I think what can be useful, is also to create `functions from closures`, where we spawn threads and instrument them, for smaller `spans`/scopes.

PR #317, opened 2024-10-06
refactor(commands): decouple logic from option structs for check, prune, repair, key, and restore

simonsan:
I saw in #224 that `CheckOptions::run()` has been moved to type `CheckResultsCollector`. In other commands we actually have functions, not associated to any type, that take their options as parameters. I applied this here to check, prune, repair, key, and restore as well. Because I think it reduces coupling and increases testability.

The idea behind having these standalone functions is, that check, prune, repair, key, and restore are not run on their options (as a method of e.g. `CheckOptions`), but rather take parameters, where one is the e.g. `CheckOption` itself.

In principle: methods that implement commands don't operate on their own options and have side effects on other types - options are passed into functions as parameters.

Furthermore, a `check()` on `CheckOptions` sounds if it validates these options itself, rather than being run on an external type. I think from that POV it also makes sense to have such freestanding functions as entry point to our commands in `rustic_core`.

### DIFF
--- a/crates/core/tests/integration.rs
+++ b/crates/core/tests/integration.rs
@@ -540,7 +540,7 @@ fn test_prune(
     let plan = repo.prune_plan(&prune_opts)?;
     // TODO: Snapshot-test the plan (currently doesn't impl Serialize)
     // assert_ron_snapshot!("prune", plan);
-    plan.do_prune(&repo, &prune_opts)?;
+    repo.prune(&prune_opts, plan)?;
 
     // run check
     let check_opts = CheckOptions::default().read_data(true);
@@ -549,7 +549,7 @@ fn test_prune(
     if !instant_delete {
         // re-run if we only marked pack files. As keep-delete = 0, they should be removed here
         let plan = repo.prune_plan(&prune_opts)?;
-        plan.do_prune(&repo, &prune_opts)?;
+        repo.prune(&prune_opts, plan)?;
         repo.check(check_opts)?;
     }
 
--- a/crates/core/src/commands/check.rs
+++ b/crates/core/src/commands/check.rs
@@ -136,106 +136,110 @@ pub struct CheckOptions {
     pub read_data_subset: ReadSubsetOption,
 }
 
-impl CheckOptions {
-    /// Runs the `check` command
-    ///
-    /// # Type Parameters
-    ///
-    /// * `P` - The progress bar type.
-    /// * `S` - The state the repository is in.
-    ///
-    /// # Arguments
-    ///
-    /// * `repo` - The repository to check
-    ///
-    /// # Errors
-    ///
-    /// If the repository is corrupted
-    pub(crate) fn run<P: ProgressBars, S: Open>(
-        self,
-        repo: &Repository<P, S>,
-        trees: Vec<TreeId>,
-    ) -> RusticResult<()> {
-        let be = repo.dbe();
-        let cache = repo.cache();
-        let hot_be = &repo.be_hot;
-        let raw_be = repo.dbe();
-        let pb = &repo.pb;
-        if !self.trust_cache {
-            if let Some(cache) = &cache {
-                for file_type in [FileType::Snapshot, FileType::Index] {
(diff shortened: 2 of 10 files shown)


crates/core/tests/integration.rs
```diff
@@ -540,7 +541,7 @@ fn test_prune(
     let plan = repo.prune_plan(&prune_opts)?;
     // TODO: Snapshot-test the plan (currently doesn't impl Serialize)
     // assert_ron_snapshot!("prune", plan);
-    plan.do_prune(&repo, &prune_opts)?;
+    prune_repository(&repo, plan, &prune_opts)?;
```
aawsome, line 544: First, we could also implement a `plan.do_prune` on `PrunePlan` and mark it as deprecated. Then we could have this PR as non-breaking.

Second, I would not make a `pub prune_repository`, but instead make this a `Repository` method. Then it would be simply `repo.prune(plan, &prune_opts)` here...

aawsome:
I agree to this refactor. Most changes are actually non-breaking as they are all offered to the users via the `Repository` API. I would suggest having this as a non-breaking change, see my comment.

Besides this, we should either postpone the topic or (preferably) merge it fast. Leaving it for some time just increases the cost of this... ;-)

crates/core/tests/integration.rs
simonsan: I agree in general, I did this, because it is relatively inconsistent over `rustic_core`, e.g. we have [pub(crate) fn copy<'a, Q, R: IndexedFull, P: ProgressBars, S: IndexedIds>(](https://github.com/rustic-rs/rustic_core/blob/f68ffa3fe1616b66b3f1b944c981155c30b86534/crates/core/src/commands/copy.rs#L44), [pub(crate) fn backup<P: ProgressBars, S: IndexedIds>(](https://github.com/rustic-rs/rustic_core/blob/f68ffa3fe1616b66b3f1b944c981155c30b86534/crates/core/src/commands/backup.rs#L214), [pub(crate) fn dump<P, S: IndexedFull>(](https://github.com/rustic-rs/rustic_core/blob/f68ffa3fe1616b66b3f1b944c981155c30b86534/crates/core/src/commands/dump.rs#L28), etc.

simonsan: So that was my intention, to make it more consistent. Each command has an entry point, and we don't implement everything on Repository, because then we get a huge master object with a lot of methods, which makes it also harder to test and couples all that functionality to that object/type. :/

simonsan: `repo.prune()` we can still implement for convenience using `prune_repository`?

simonsan: Implemented `prune()` on repo, which uses `prune_repository()`.

simonsan:
It's true, all the changes are non-breaking now. I added the `deprecation` notes as well.

I made single commits for each command being decoupled and force pushed, so it's easier to review.

crates/core/tests/integration.rs
```diff
@@ -540,7 +540,7 @@ fn test_prune(
     let plan = repo.prune_plan(&prune_opts)?;
     // TODO: Snapshot-test the plan (currently doesn't impl Serialize)
     // assert_ron_snapshot!("prune", plan);
-    plan.do_prune(&repo, &prune_opts)?;
+    repo.prune(&prune_opts, plan)?;
```
simonsan, line 543: I migrated this, although the old API should still work.

crates/core/tests/integration.rs
```diff
@@ -549,7 +549,7 @@ fn test_prune(
     if !instant_delete {
         // re-run if we only marked pack files. As keep-delete = 0, they should be removed here
         let plan = repo.prune_plan(&prune_opts)?;
-        plan.do_prune(&repo, &prune_opts)?;
+        repo.prune(&prune_opts, plan)?;
```
simonsan, line 552: I migrated this, although the old API should still work.

aawsome (approved):

PR #318, opened 2024-10-06
Draft: [EXPERIMENT] Sketching out a RusticIssueCollector

simonsan:
Implementation to collect feedback and sketch out an idea. We should iterate more on it, to make it more ergonomic, I think.

### DIFF
--- a/crates/core/src/error/collector.rs
+++ b/crates/core/src/error/collector.rs
@@ -6,13 +6,13 @@ use log::{error, info, warn};
 /// A rustic issue result
 ///
 /// rustic issue results are used to return a result along with possible issues.
-pub type RusticIssueResult<T> = Result<(T, Option<Vec<RusticIssue>>), Vec<RusticError>>;
+pub(crate) type RusticIssueResult<T> = Result<(T, Option<Vec<RusticIssue>>), Vec<RusticError>>;
 
 /// A rustic issue
 ///
 /// An issue is a message that can be logged to the user.
 #[derive(Debug)]
-pub enum RusticIssue {
+pub(crate) enum RusticIssue {
     /// An error issue, indicating that something went wrong irrecoverably
     Error(RusticError),
 
@@ -49,7 +49,7 @@ impl RusticIssue {
 ///
 /// Warning messages are used to indicate that something might be wrong.
 #[derive(Debug, Clone, From)]
-pub struct RusticWarning(String);
+pub(crate) struct RusticWarning(String);
 
 impl RusticWarning {
     pub fn new(message: &str) -> Self {
@@ -67,7 +67,7 @@ impl From<&str> for RusticWarning {
 ///
 /// Info messages are used to provide additional information to the user.
 #[derive(Debug, Clone, From)]
-pub struct RusticInfo(String);
+pub(crate) struct RusticInfo(String);
 
 impl RusticInfo {
     pub fn new(message: &str) -> Self {
@@ -82,7 +82,7 @@ impl From<&str> for RusticInfo {
 }
 
 #[derive(Debug, Default)]
-pub struct RusticIssueCollector {
+pub(crate) struct RusticIssueCollector {
     /// The errors collected
     errors: Option<Vec<RusticError>>,
 
--- a/crates/core/src/error/collector.rs
+++ b/crates/core/src/error/collector.rs
@@ -24,19 +24,19 @@ pub(crate) enum RusticIssue {
 }
 
 impl RusticIssue {
-    pub fn new_error(error: RusticErrorKind) -> Self {
+    pub(crate) fn new_error(error: RusticErrorKind) -> Self {
         Self::Error(error.into())
     }
 
-    pub fn new_warning(message: &str) -> Self {
+    pub(crate) fn new_warning(message: &str) -> Self {
         Self::Warning(message.into())
     }
 
-    pub fn new_info(message: &str) -> Self {
+    pub(crate) fn new_info(message: &str) -> Self {
         Self::Info(message.into())
     }
 
-    pub fn log(&self) {
+    pub(crate) fn log(&self) {
         match self {
             Self::Error(error) => error!("{}", error),
             Self::Warning(warning) => warn!("{}", warning.0),
@@ -52,7 +52,7 @@ impl RusticIssue {
 pub(crate) struct RusticWarning(String);
 
 impl RusticWarning {
-    pub fn new(message: &str) -> Self {
+    pub(crate) fn new(message: &str) -> Self {
         Self(message.to_owned())
     }
 }
@@ -70,7 +70,7 @@ impl From<&str> for RusticWarning {
 pub(crate) struct RusticInfo(String);
 
 impl RusticInfo {
-    pub fn new(message: &str) -> Self {
+    pub(crate) fn new(message: &str) -> Self {
         Self(message.to_owned())
     }
 }
@@ -97,7 +97,7 @@ pub(crate) struct RusticIssueCollector {
 }
 
 impl RusticIssueCollector {
-    pub fn new(log: bool) -> Self {
+    pub(crate) fn new(log: bool) -> Self {
         Self {
             errors: None,
             warnings: None,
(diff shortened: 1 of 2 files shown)


crates/core/src/error/collector.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// The warnings collected
+    warnings: Option<Vec<RusticWarning>>,
+
+    /// The info collected
+    info: Option<Vec<RusticInfo>>,
+
+    /// Whether to log items directly during addition
+    log: bool,
+}
```
simonsan, line 104: We might want to adapt it, so we can easier use it from multiple threads, and other contexts, e.g. via Mutexes or Channels.

crates/core/src/error/collector.rs
```diff
@@ -0,0 +1,289 @@
...
+    Err(Vec<RusticError>),
+}
+
+/// A rustic issue
+///
+/// An issue is a message that can be logged to the user.
+#[derive(Debug)]
+pub enum RusticIssue {
+    /// An error issue, indicating that something went wrong irrecoverably
+    Error(RusticError),
```
simonsan, line 24: We might not have that within this Enum though? 🤔  as that should always be a hard error I think? Not sure though.

crates/core/src/error/collector.rs
```diff
@@ -0,0 +1,289 @@
...
+///
+/// rustic issue results are used to return a result along with possible issues.
+#[derive(Debug)]
+pub enum RusticIssueResult<T> {
+    /// For when we return a result along with possible issues
+    Ok(T, Option<Vec<RusticIssue>>),
+
+    /// For when we abort and return only the errors that occurred
+    Err(Vec<RusticError>),
+}
```
aawsome, line 16: This is actually equivalent to `Result<(T, Option<Vec<RusticIssue>>), Vec<RusticError>>`

Besides this, I would not us `Option<Vec<..>>` but only `Vec<..>` and treat an empty `Vec` like the `None` case. I think this simplifies implementation

crates/core/src/error/collector.rs
aawsome: Actually I do think that we can have Errors on a deeper level which are shown as warnings on a higher level. Something like backup failed for creating the first snapshot, but we want to continue creating the second one...

crates/core/src/error/collector.rs
```diff
@@ -0,0 +1,289 @@
...
+
+/// A rustic issue
+///
+/// An issue is a message that can be logged to the user.
+#[derive(Debug)]
+pub enum RusticIssue {
```
aawsome, line 22: Why not using something like `error: RusticError` and `severity: RusticSeverity` with the latter being an enum? I can imagine we can have the exactly identical error which is at some place classified as warning (or even info) and at another place just an error.

crates/core/src/error/collector.rs
```diff
@@ -0,0 +1,289 @@
...
+            Self::Info(info) => info!("{}", info.0),
+        }
+    }
+}
+
+/// A rustic warning message
+///
+/// Warning messages are used to indicate that something might be wrong.
+#[derive(Debug, Clone, From)]
+pub struct RusticWarning(String);
```
aawsome, line 59: I don't like having `String` here. IMO we should just have `RusticError`

aawsome (commented):
Besides some specific comments, a general one:

As it seems there is no crate yet to provide such a functionality we need, I propose we make a new crate out of this which we then use within rustic_core and rustic...

I could even think of having some macro functionality to simplify writing functions which collect errors without the need to explicitly create a `Collector` for collecting errors. Something like
```rust
#[collect]
fn my_func() -> CollectorResult<T> {
  let x = fn1()? // returns CollectorResult; on Err just bubble this up
          .fn2().collect_as_err_or(0.0, MyErr);  // returns CollectorResult, on error collect errors as error and use value 0.0
  let y = fn3(x).collect_as_warn_or(1.0, MyWarn); // as before, but warning
  Ok(x)
}
```

crates/core/src/error/collector.rs
simonsan: I don't know, I would not classify an error at one place as a warning and at the other as an info. An error is an error, and should be clearly communicated as one. If we change the general meaning of an error, then it shouldn't be an error in the first place.

simonsan: But I do know what you mean, something like: 'File not found'. At one place, it can be irrecoverable, e.g. a config file not found for a backend. At some other place, it maybe can be continued with default settings.

But in that case we should handle that error in-place and not return an error I feel.

crates/core/src/error/collector.rs
simonsan: In this case, we should handle that at the point we get that `error` and start the backup for the second `path` for example. The failed backup for the first will be output to the user at the `match` of the result (e.g. in a loop) and we start the next one, but give user feedback.

simonsan: So I think this is the same topic as above, we shouldn't give `errors` different meanings, I don't know if they are really contextual in that sense. There can be hard and soft errors, for sure. But I think we are going to replace the term soft error here, with an `issue` which can result in a `warning` or `info`. But then we need to reserve the term `error` for things that are irrecoverable.

crates/core/src/error/collector.rs
simonsan: Same as above, this here is a `warning`, `RusticError` should be something that we return to the user, in case we can't recover from something. This is pub here, but I'm not sure if we ever should make RusticWarning, RusticInfo part of our public API. Users of `rustic_core`, also `rustic-rs` should never have to deal with that. They will only get `errors` from us. I think 🤔 If it is recoverable, then we should recover from it, not the users.

crates/core/src/error/collector.rs
simonsan: Yes, still a draft and I was sketching out that type and landed there 😅

crates/core/src/error/collector.rs
simonsan: Also, important to say here: `RusticIssue`, `RusticWarning`, and `RusticIssueCollector` should never go over the library boundary. So it's for internal use only, it should not be part of our public API, and we should actually not return `RusticIssue` from any public function. It should only be used for internal error handling, so we as library authors can judge situations better internally, and give our users errors only in cases that are worthwhile.

Hence, I reduced the visibility of all the types in this file to `pub(crate)`

simonsan:
I think one of our problems is the thought, that we would need to transfer warnings and information over a function boundary and not handle them as locally as possible.

From this thought stems the problematic, that we want to have `warnings` and `info` and collect them into a `RusticIssue`. Maybe it's worth to reconsider that and use the try-`?`-operator less, and handle everything as it comes without bubbling it up, too much. We probably want to check out our error paths at that point.

Some key insights:

- overusing the propagation of `warnings` and `info` across function boundaries. This is leading to complex error handling, where we're going to be collecting and returning various levels of issues (Error, Warning, Info) as part of the overall function result. This complexity can make it difficult to manage and reason about errors.

- whenever a warning or informational message occurs, we should handle it immediately rather than pushing it up the call stack (locality)

- we should reserve the propagation of errors (?, or using Result) for cases where a failure should affect the calling function, overusing them can lead to over-bubbling of non-critical issues up the call stack

- We should keep our Result<T, RusticError> focused on representing success and failure states, warnings and informational messages shouldn’t affect the primary result

- replacing `log` with `tracing` is a core part of that strategy to introduce structured logging so we are easier able to determine the core error flow

simonsan:
From this point, there shouldn't be actually any `CollectorResult`, as that would pass the function boundary. Which would be counterproductive to our efforts. Such a Collector should be local to the function itself, and we should only return values out of it, e.g. collected `Errors`. And spit out the relevant `warnings` and `infos` at the end **before** returning with `::log_all`.

simonsan:
The more I think about and work with it, the less I'm sure, that this is a valid approach for `rustic_core`.

Especially, when it comes to just pushing a lot of errors up the call stack. I do see a use case for `check` function for that, but there we could return a `Result<T, Vec<RusticError>>` or a `Result<T, HashMap<SeverityLevel, Vec<RusticError>>` and not have a complete structure, just about prolonging error handling.

The problem I have with `SeverityLevel` is, that `Result<T, Err>` essentially already does that. `T` can be already returning 'Something is strange and should be corrected, but repository integrity is not affected.', while `Err` is already 'Something in the repository is not as it should be.'. 

Could we do it like this: We merge https://github.com/rustic-rs/rustic_core/pull/317 (otherwise we need to rebase a lot of stuff) and then I would like to try to implement a proposal on top of that of how I could imagine handling errors in `check`.

I'm closing this as an experiment for now.

PR #321, opened 2024-10-07
refactor(errors)!: improve RusticError display and usage

simonsan:
TODO:

- [x] fix `todo!("Error transition")`
- [x] check and group imports `(ErrorKind, RusticError, RusticResult)`
- [x] fix tests
- [x] fix clippy
- [x] fix build for *nix and osx
- [x] check and adapt documentation of errors in functions
- [x] remove variants from local errors that are now unused
- [x] build `rustic-rs` against this branch and check, that everything works
- [x] check `ErrorKind::Command` variant and usage, probably not needed variant, as most of it is also `Internal` -> replace and remove variant

### DIFF
--- a/crates/backend/src/local.rs
+++ b/crates/backend/src/local.rs
@@ -108,7 +108,10 @@ impl LocalBackend {
     ///
     /// # Errors
     ///
-    // TODO: Add error types
+    /// * If the patterns could not be compiled.
+    /// * If the command could not be parsed.
+    /// * If the command could not be executed.
+    /// * If the command was not successful.
     ///
     /// # Notes
     ///
@@ -321,7 +324,10 @@ impl ReadBackend for LocalBackend {
     ///
     /// # Errors
     ///
-    // TODO: Add error types
+    /// * If the file could not be opened.
+    /// * If the file could not be sought to the given position.
+    /// * If the length of the file could not be converted to u32.
+    /// * If the exact length of the file could not be read.
     fn read_partial(
         &self,
         tpe: FileType,
@@ -429,7 +435,11 @@ impl WriteBackend for LocalBackend {
     ///
     /// # Errors
     ///
-    // TODO: Add error types
+    /// * If the file could not be opened.
+    /// * If the length of the bytes could not be converted to u64.
+    /// * If the length of the file could not be set.
+    /// * If the bytes could not be written to the file.
+    /// * If the OS Metadata could not be synced to disk.
     fn write_bytes(
         &self,
         tpe: FileType,
--- a/crates/backend/src/local.rs
+++ b/crates/backend/src/local.rs
@@ -405,7 +405,7 @@ impl WriteBackend for LocalBackend {
         trace!("creating repo at {:?}", self.path);
         fs::create_dir_all(&self.path).map_err(|err| {
             RusticError::with_source(
-                ErrorKind::Io,
+                ErrorKind::InputOutput,
                 "Failed to create the directory. Please check the path and try again.",
                 err,
             )
@@ -416,7 +416,7 @@ impl WriteBackend for LocalBackend {
             let path = self.path.join(tpe.dirname());
             fs::create_dir_all(path.clone()).map_err(|err| {
                 RusticError::with_source(
-                    ErrorKind::Io,
+                    ErrorKind::InputOutput,
                     "Failed to create the directory. Please check the path and try again.",
                     err,
                 )
@@ -428,7 +428,7 @@ impl WriteBackend for LocalBackend {
             let path = self.path.join("data").join(hex::encode([i]));
             fs::create_dir_all(path.clone()).map_err(|err| {
                 RusticError::with_source(
-                    ErrorKind::Io,
+                    ErrorKind::InputOutput,
                     "Failed to create the directory. Please check the path and try again.",
                     err,
                 )
@@ -471,7 +471,7 @@ impl WriteBackend for LocalBackend {
             .open(&filename)
             .map_err(|err| {
                 RusticError::with_source(
-                    ErrorKind::Io,
+                    ErrorKind::InputOutput,
                     "Failed to open the file. Please check the file and try again.",
                     err,
                 )
@@ -485,7 +485,7 @@ impl WriteBackend for LocalBackend {
         })?)
         .map_err(|err| {
             RusticError::with_source(
-                ErrorKind::Io,
+                ErrorKind::InputOutput,
                 "Failed to set the length of the file. Please check the file and try again.",
                 err,
             )
@@ -494,7 +494,7 @@ impl WriteBackend for LocalBackend {
 
         file.write_all(&buf).map_err(|err| {
             RusticError::with_source(
-                ErrorKind::Io,
+                ErrorKind::InputOutput,
                 "Failed to write to the buffer. Please check the file and try again.",
                 err,
             )
@@ -503,7 +503,7 @@ impl WriteBackend for LocalBackend {
 
         file.sync_all().map_err(|err| {
             RusticError::with_source(
-                ErrorKind::Io,
+                ErrorKind::InputOutput,
                 "Failed to sync OS Metadata to disk. Please check the file and try again.",
                 err,
             )
--- a/crates/backend/src/local.rs
+++ b/crates/backend/src/local.rs
@@ -226,29 +226,26 @@ impl ReadBackend for LocalBackend {
 
         if tpe == FileType::Config {
             return Ok(if path.exists() {
-                vec![(
-                    Id::default(),
-                    path.metadata()
-                        .map_err(|err|
+                vec![(Id::default(), {
+                    let metadata = path.metadata().map_err(|err|
                             RusticError::with_source(
                                 ErrorKind::Backend,
                                 "Failed to query metadata of the file. Please check the file and try again.",
                                 err
                             )
                             .attach_context("path", path.to_string_lossy())
-                        )?
-                        .len()
-                        .try_into()
-                        .map_err(|err|
-                            RusticError::with_source(
-                                ErrorKind::Backend,
-                                "Failed to convert file length to u32.",
-                                err
-                            )
-                            .attach_context("length", path.metadata().unwrap().len().to_string())
-                            .ask_report()
-                        )?,
-                )]
+                        )?;
+
+                    metadata.len().try_into().map_err(|err| {
+                        RusticError::with_source(
+                            ErrorKind::Backend,
+                            "Failed to convert file length to u32.",
+                            err,
+                        )
+                        .attach_context("length", metadata.len().to_string())
+                        .ask_report()
+                    })?
+                })]
             } else {
                 Vec::new()
             });
@@ -266,7 +263,8 @@ impl ReadBackend for LocalBackend {
             .map(|e| -> RusticResult<_> {
                 Ok((
                     e.file_name().to_string_lossy().parse()?,
-                    e.metadata()
+                    {
+                        let metadata = e.metadata()
                         .map_err(|err|
                             RusticError::with_source(
                                 ErrorKind::Backend,
@@ -275,7 +273,9 @@ impl ReadBackend for LocalBackend {
                             )
                             .attach_context("path", e.path().to_string_lossy())
                         )
-                        ?
+                        ?;
+
+                        metadata
                         .len()
                         .try_into()
                         .map_err(|err|
@@ -284,9 +284,10 @@ impl ReadBackend for LocalBackend {
                                 "Failed to convert file length to u32.",
                                 err
                             )
-                            .attach_context("length", e.metadata().unwrap().len().to_string())
+                            .attach_context("length", metadata.len().to_string())
                             .ask_report()
-                        )?,
+                        )?
+                    },
                 ))
             })
             .inspect(|r| {
--- a/crates/backend/src/local.rs
+++ b/crates/backend/src/local.rs
@@ -1,9 +1,8 @@
 use std::{
     fs::{self, File},
     io::{Read, Seek, SeekFrom, Write},
-    num::TryFromIntError,
     path::{Path, PathBuf},
-    process::{Command, ExitStatus},
+    process::Command,
 };
 
 use aho_corasick::AhoCorasick;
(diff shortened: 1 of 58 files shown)


aawsome:
I think this is mixing two things which should be handled completely independent from each other: Channels and error propagation.

Actually, we don't need a channel if we want to collect any kind of information (be it an array of errors, a bool `isErr` or whatever else) from a called function/method. For this the return value (or maybe a `&mut` argument) suffices, is simpler and more performant.
We only need a channel to communicate (i.e. send information and this can be an array of errors, a bool or any other result) between threads, i.e. if we spawn threads or use already-spawned threads to do some work.

To be more precise, in this example we could change `fn xy(...,err_send: Sender<CheckErrorKind>) -> Result<T>` into `fn xy(..., is_err: &AtomicBool) -> Result<T>` and directly modify `is_err` or even use `fn xy(...) -> Result<(T,bool)>` and propagate `is_err`. This gives exactly the same functionality without using a channel.

IMO there is just one fundamental decision to make: Do we want to return and propagate an error flag or a list of errors?

aawsome:
And the same applies to warning, where we also have to decide:
- do we want to just log them and not give the information to the caller?
- do we want to give a `is_warn` bool information to the caller?
- do we want to give a list of warnings to the caller?

simonsan:
> I think this is mixing two things which should be handled completely independent from each other: Channels and error propagation.
> 
> Actually, we don't need a channel if we want to collect any kind of information (be it an array of errors, a bool `isErr` or whatever else) from a called function/method. For this the return value (or maybe a `&mut` argument) suffices, is simpler and more performant. We only need a channel to communicate (i.e. send information and this can be an array of errors, a bool or any other result) between threads, i.e. if we spawn threads or use already-spawned threads to do some work.
> 
> To be more precise, in this example we could change `fn xy(...,err_send: Sender<CheckErrorKind>) -> Result<T>` into `fn xy(..., is_err: &AtomicBool) -> Result<T>` and directly modify `is_err` or even use `fn xy(...) -> Result<(T,bool)>` and propagate `is_err`. This gives exactly the same functionality without using a channel.
> 
> IMO there is just one fundamental decision to make: Do we want to return and propagate an error flag or a list of errors?

I think collecting all errors and returning them from the `check_repository` function makes sense, as we are processing a collection of data with multiple ways to fail. `Repository::check()` itself though, should only return a single error stating that the check failed imho.

I agree, I used the channel here, because I got confused by the parallel iterator. But instead of a `for_each`, we can just use `map` and `collect` the results. Then we can e.g. partition the results and return a `RusticResult<Result<() | Vec<RusticWarning>, Vec<RusticError>>>`. Where the inner Result is for soft-errors/warnings (that would result in a warning or just a logged error, depending on its severity that we would determine in the callers `match`). And the outer result a hard-error.

The `RusticWarning` here should stay internal use only, and never propagate to the user of the lib, it's only for logging purposes and further handling - I would even think, we don't even need it and could log a warning directly where it is important. So we can easily find the spot where the warning was emitted. The `Vec<RusticError>` should also not become part of the Repository API as the users should not need to handle multiple errors and decide on the severity on their own - for them the `RusticResult<()>` with a single error should be the contract. So they can rely on either the check being ok, or the check having failed.

simonsan (commented):
Looked through myself again (more than once), and this is what I'm still unsure about.

crates/core/src/commands/restore.rs
```diff
@@ -537,14 +555,7 @@ fn restore_contents<P: ProgressBars, S: Open>(
...
                                 if filesize > 0 {
-                                    dest.set_length(path, filesize)
-                                        .map_err(|err| {
-                                            CommandErrorKind::ErrorSettingLength(
-                                                path.clone(),
-                                                Box::new(err),
-                                            )
-                                        })
-                                        .unwrap();
+                                    dest.set_length(path, filesize).unwrap();
```
nardoor, line 558: could this be the opportunity to remove these `unwrap`?

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-};
+/// Severity of an error, ranging from informational to fatal.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Severity {
+    /// Informational
+    Info,
 
-/// Result type that is being returned from methods that can fail and thus have [`RusticError`]s.
-pub type RusticResult<T> = Result<T, RusticError>;
+    /// Warning
```
nardoor, line 73: two questions about `Severity` : 
- is `Info` really a possible severity for our errors?
- `Severity` seems to never be used in this refactor (apart from re-exporting), should we keep it?

About both these questions I feel like `RusticError` is overall a non-recoverable error:
- we have nested `Result`s for recoverable errors
- `RusticError` is not made to be programmatically dealt with, it is an error to the user. You cannot `match` a rustic error more than the rather large `ErrorKind` enum.

For this reason I feel like a `RusticError` is always `Error` if not `Fatal`.

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-    /// Checks if the error is due to an incorrect password
-    pub fn is_incorrect_password(&self) -> bool {
-        matches!(
-            self.0,
-            RusticErrorKind::Repository(RepositoryErrorKind::IncorrectPassword)
-        )
-    }
+/// Status of an error, indicating whether it is permanent, temporary, or persistent.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Status {
```
nardoor, line 85: just a note: not yet used

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-/// [`ConfigFileErrorKind`] describes the errors that can be returned for `ConfigFile`s
-#[derive(Error, Debug, Display)]
-pub enum ConfigFileErrorKind {
-    /// config version not supported!
-    ConfigVersionNotSupported,
-    /// Parsing Polynomial in config failed: `{0:?}`
-    ParsingFailedForPolynomial(#[from] ParseIntError),
-}
+    /// Severity of the error.
+    severity: Option<Severity>,
```
nardoor, line 170: as per my former comment:
`Severity` to me is always `Error` if not `Fatal`

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-    PartialReadOfPackfileFailed,
-    /// writing Bytes failed
-    WritingBytesFailed,
-    /// `{0}`
-    #[error(transparent)]
-    PackDecryptionFailed(#[from] CryptBackendErrorKind),
+    /// Backtrace of the error.
+    ///
+    // Need to use option, otherwise thiserror will not be able to derive the Error trait.
+    backtrace: Option<Backtrace>,
```
nardoor, line 178: sad it's `Option`al but as we deal internally with it, `rustic_core` developers can make sure it's never put as `None`.

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-    /// Specify one of the keep-* options for forget! Please use keep-none to keep no snapshot.
-    NoKeepOption,
-    /// `{0}`
-    #[error(transparent)]
-    FromParseError(#[from] shell_words::ParseError),
-}
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+/// Errors that can result from rustic.
+pub struct RusticError {
```
nardoor, line 144: What do you think about ordering the `fields` in the same order as we use them in the `Display` implementation?

nardoor: To me this could make it "easier" to make sure all the fields are used in the `impl Disaply`

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
+impl RusticError {
+    /// Creates a new error with the given kind and guidance.
+    pub fn new(kind: ErrorKind, guidance: impl Into<SmolStr>) -> Box<Self> {
+        Box::new(Self {
+            kind,
+            guidance: guidance.into(),
+            context: Cow::default(),
+            source: None,
+            error_code: None,
+            docs_url: None,
```
nardoor, line 249: constructor code is "duplicated" several times.
would there be a way to reduce that without cost ? 

`Default`? `builder pattern`? `...spread operator`?

crates/core/src/error.rs
```diff
@@ -1,810 +1,451 @@
...
-    fn from(value: E) -> Self {
-        Self(RusticErrorKind::from(value))
+// Setters for anything we do want to expose publicly.
+//
+// These were initially generated by `derive_setters`,
+// and then manually adjusted to return `Box<Self>` instead of `Self` which
+// unfortunately is not possible with the current version of the `derive_setters`.
+//
+// BEWARE! `attach_context` is manually implemented to allow for multiple contexts
+// to be added and is not generated by `derive_setters`.
```
nardoor, line 325: I don't see `derive_setters` (I didn't know about that derive), where is it used?

nardoor (commented):
Hello, overall I think this refactor is a nice and large step in a user-oriented direction.

As the volume of the PR and the time I had, I reviewed `local.rs` and quickly `opendal.rs` to see it in action. Then I focused on the `error.rs` definitions.

I'll continue later this week to go through other files.

Even thought I didn't finish reading I can already say good job on getting so much work done in rather little time.

crates/backend/src/local.rs
```diff
@@ -169,6 +203,11 @@ impl ReadBackend for LocalBackend {
 
         let walker = WalkDir::new(self.path.join(tpe.dirname()))
             .into_iter()
+            .inspect(|r| {
+                if let Err(err) = r {
+                    error!("Error while listing files: {err:?}");
+                }
+            })
```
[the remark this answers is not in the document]
simonsan: Yes, I think special error handling cases need to come a bit later, because otherwise this gets too big. I actually introduced this, because before we just threw away the error, doing nothing.

crates/core/src/error.rs
simonsan: Actually in a lot of places, it replaces the builder pattern for us, in a way.

See `rg 'use derive_setters::'`

crates/backend/src/local.rs
```diff
@@ -259,37 +339,59 @@ impl ReadBackend for LocalBackend {
...
-    /// * [`LocalBackendErrorKind::ReadingExactLengthOfFileFailed`] - If the length of the file could not be read.
-    ///
-    /// [`LocalBackendErrorKind::OpeningFileFailed`]: LocalBackendErrorKind::OpeningFileFailed
-    /// [`LocalBackendErrorKind::CouldNotSeekToPositionInFile`]: LocalBackendErrorKind::CouldNotSeekToPositionInFile
-    /// [`LocalBackendErrorKind::FromTryIntError`]: LocalBackendErrorKind::FromTryIntError
-    /// [`LocalBackendErrorKind::ReadingExactLengthOfFileFailed`]: LocalBackendErrorKind::ReadingExactLengthOfFileFailed
+    /// * If the file could not be opened.
+    /// * If the file could not be sought to the given position.
+    /// * If the length of the file could not be converted to u32.
+    /// * If the exact length of the file could not be read.
```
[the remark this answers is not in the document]
simonsan: Agreed!

crates/backend/src/local.rs
```diff
@@ -328,36 +450,70 @@ impl WriteBackend for LocalBackend {
...
+            })?;
+
+        file.set_len(buf.len().try_into().map_err(|err| {
+            RusticError::with_source(
+                ErrorKind::Internal,
+                "Failed to convert length to u64. This is a bug. Please report it.",
+                err,
+            )
+            .attach_context("length", buf.len().to_string())
+        })?)
```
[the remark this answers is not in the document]
simonsan: Yeah, but I think this is a small cost, if we get better errors for it long-term and guidance for the users.

crates/core/src/commands/restore.rs
simonsan: I want to do that at a later stage, when we can focus on fixing the unwraps itself. This PR's focus should really be the RusticResult/RusticError, local errors and `map_err`, IMHO.

crates/core/src/error.rs
simonsan: Yeah, severity might be useless overall. As well as status maybe. I was thinking of having it, so we can use it down the line without breaking our API for these things. Though that shouldn't be the case, now that we use `non_exhaustive`. So for sure, we can remove that.

crates/core/src/error.rs
simonsan: Yes, same as severity. I was unsure where to use it. We can also remove it. If we think we may never use it.

crates/core/src/error.rs
simonsan: Yeah, but sometimes, we don't have a source, e.g. when coming from `ok_or_else`

crates/core/src/error.rs
simonsan: Thanks for the reminder, I did that during sketching and actually forgot!

crates/backend/src/opendal.rs
```diff
@@ -183,35 +264,94 @@ impl ReadBackend for OpenDALBackend {
...
+            .map(|e| -> RusticResult<(Id, u32)> {
+                Ok((
+                    e.name().parse()?,
+                    e.metadata()
+                        .content_length()
+                        .try_into()
+                        .map_err(|err|
+                            RusticError::with_source(
+                                ErrorKind::Internal,
+                                "Parsing content length failed",
```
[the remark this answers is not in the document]
simonsan: I can't find it, actually D:

crates/core/src/error.rs
simonsan: That being said, we might want others also to construct errors, e.g. when they implement some backend related things. So we may need different status and severity information there. Maybe not the one, I took for now. E.g. informational might not be useful for errors. But maybe useful for `RusticIssues`, the ones we want to deal with for `CheckCommand`.

crates/core/tests/snapshots/errors__error_display.snap
```diff
@@ -0,0 +1,41 @@
...
+If you can, please attach an anonymized debug log to the issue.
+
+Thank you for helping us improve rustic!
+
+
+Some additional details ...
+
+Context:
+- path:	/path/to/file
+- called:	used s3 backend
```
simonsan, line 31: Probably not worth to have a `\t` here in between `key` and `value`. Should remove it, probably doesn't need to be aligned.

crates/core/tests/snapshots/errors__error_display.snap
```diff
@@ -0,0 +1,41 @@
...
+
+We believe this is a bug, please report it by opening an issue at:
+https://github.com/rustic-rs/rustic_core/issues/new
+
+If you can, please attach an anonymized debug log to the issue.
+
+Thank you for helping us improve rustic!
+
+
+Some additional details ...
```
simonsan, line 27: Something like

```suggestion
### Some additional details ###
```

crates/core/tests/snapshots/errors__error_display.snap
```diff
@@ -0,0 +1,41 @@
...
+
+Caused by:
+Networking Error : (source: None)
+
+Severity: Error
+
+Status: Permanent
+
+Backtrace:
+disabled backtrace
```
simonsan, line 41: Either it should not be printed, or we should give some hint, how to enable it, when it's disabled.

crates/core/src/commands/cat.rs
```diff
@@ -112,9 +104,13 @@ pub(crate) fn cat_tree<P: ProgressBars, S: IndexedTree>(
         &repo.pb.progress_counter("getting snapshot..."),
     )?;
     let node = Tree::node_from_path(repo.dbe(), repo.index(), snap.tree, Path::new(path))?;
-    let id = node
-        .subtree
-        .ok_or_else(|| CommandErrorKind::PathIsNoDir(path.to_string()))?;
+    let id = node.subtree.ok_or_else(|| {
+        RusticError::new(
+            ErrorKind::Command,
```
simonsan, line 109: Should be I/O probably

crates/core/src/commands/check.rs
```diff
@@ -667,7 +687,18 @@ fn check_pack(
 
     // check header length
     let header_len = PackHeaderRef::from_index_pack(&index_pack).size();
-    let pack_header_len = PackHeaderLength::from_binary(&data.split_off(data.len() - 4))?.to_u32();
+    let pack_header_len = PackHeaderLength::from_binary(&data.split_off(data.len() - 4))
+        .map_err(|err| {
+            RusticError::with_source(
+                ErrorKind::Command,
```
simonsan, line 693: Probably `Internal` is better here. Not sure, if it even makes sense to have the `Command` variant. As most of that is also internal.

crates/core/src/commands/check.rs
```diff
@@ -676,7 +707,17 @@ fn check_pack(
     // check header
     let header = be.decrypt(&data.split_off(data.len() - header_len as usize))?;
 
-    let pack_blobs = PackHeader::from_binary(&header)?.into_blobs();
+    let pack_blobs = PackHeader::from_binary(&header)
+        .map_err(|err| {
+            RusticError::with_source(
+                ErrorKind::Command,
```
simonsan, line 713: Same: Probably `Internal` is better here. Not sure, if it even makes sense to have the `Command` variant. As most of that is also internal.

PR #322, opened 2024-10-08
refactor(errors): improve displaying of errors

### LINKED ISSUE #231: error: data decryption failed
I get the following:
```
$ rustic -P slim -P Pcloud backup                                                                                                             
[INFO] using config /home/sperber/.config/rustic/slim.toml
[INFO] using config /home/sperber/.config/rustic/Pcloud.toml
[INFO] repository rclone:Pcloud:Rustic: password is correct.
[INFO] using cache at /home/sperber/.cache/rustic/f35f1623c65f6e8ed0424d28afc17d88d13bc4dc1687c768c732d4ec1403ccbd
error: data decryption failed
```

If i delete the local file and run the command again, it first downloads something from the remote server but then fails with exactly the same output.

Is there something i can do? Can `rustic` e.g. go back to the last "good" entry point instead of failing immediately?

simonsan:
Fixes #231

### DIFF
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -175,69 +175,74 @@ pub enum RusticErrorKind {
 /// [`CommandErrorKind`] describes the errors that can happen while executing a high-level command
 #[derive(Error, Debug, Display)]
 pub enum CommandErrorKind {
-    /// path is no dir: `{0:?}`
+    /// path is no dir: `{0}`
     PathIsNoDir(String),
-    /// used blobs are missing: blob {0} doesn't existing
+    /// used blobs are missing: blob `{0}` doesn't existing
     BlobsMissing(BlobId),
-    /// used pack {0}: size does not match! Expected size: {1}, real size: {2}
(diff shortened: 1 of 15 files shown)


crates/core/src/error.rs
```diff
@@ -175,69 +175,74 @@ pub enum RusticErrorKind {
...
+    /// [`shell_words::ParseError`]
+    #[error(transparent)]
     FromParseError(#[from] shell_words::ParseError),
 }
 
 /// [`CryptoErrorKind`] describes the errors that can happen while dealing with Cryptographic functions
 #[derive(Error, Debug, Display, Copy, Clone)]
 pub enum CryptoErrorKind {
-    /// data decryption failed
+    /// data decryption failed: `{0:?}`
```
simonsan, line 242: The underlying error was not shown to the user, so this ended in a simple: `data decryption failed`. Which is not really helpful. The hope is, that `aead:Error` is a bit more helpful here.

simonsan:
Was merged with #323

PR #334, opened 2024-10-14
refactor(error): improve error messages and file handling

### LINKED ISSUE #310: Uncaught rclone errors causing rustic panic
From time to time I experience `rustic` panic. It is always with repositories accessed using `rclone`. Example below:

```
$ cat /var/folders/36/02d6s6313zq5k3mfrhg4d8zc0000gn/T/report-75c43922-e59d-44ef-8ef7-ff00ce720e32.toml
name = "rustic-rs"
operating_system = "Mac OS 15.0.1 [64-bit]"
crate_version = "0.9.1"
explanation = """
Panic occurred in file '/Users/kptsky/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustic_core-0.5.1/src/commands/check.rs' at line 229
"""
cause = '''
called `Result::unwrap()` on an `Err` value: backoff failed: Permanent(reqwest::Error { kind: Status(404), url: "http://127.0.0.1:52633/data/3130f02f6442ac6b053d2b2ffce29b5263d30e0897385be796095cd5510be9bb" })

Caused by:
    HTTP status client error (404 Not Found) for url (http://127.0.0.1:52633/data/3130f02f6442ac6b053d2b2ffce29b5263d30e0897385be796095cd5510be9bb)'''
method = "Panic"
backtrace = """

   0: 0x108d7cd8a - __mh_execute_header
   1: 0x108d7d285 - __mh_execute_header
   2: 0x1088d5572 - __mh_execute_header
   3: 0x1088d5ab6 - __mh_execute_header
   4: 0x1088d5502 - __mh_execute_header
   5: 0x1088d5d82 - __mh_execute_header
   6: 0x108d9322d - __mh_execute_header
   7: 0x10803858c - __mh_execute_header
   8: 0x10803834b - __mh_execute_header
   9: 0x108bedf1b - __mh_execute_header
  10: 0x7ff81a603253 - __pthread_start"""
```

It is always `HTTP status client error (404 Not Found) for url (http://127.0.0.1:xxxxx/`. 

Definitely nothing critical but would be nice just to log an error instead of pressing panic button:)

simonsan:
- Create parent directory if it does not exist before opening the file for writing.

Fixes https://github.com/rustic-rs/rustic/issues/1315
Fixes #310


simonsan:
I kept:

```rust
        for i in 0u8..=255 {
            let filename = self.path.join("data").join(hex::encode([i]));
            fs::create_dir_all(&filename).map_err(|err| {
                LocalBackendErrorKind::DirectoryCreationFailed {
                    path: filename.clone(),
                    source: err,
                }
            })?;
        }
```

as that impacts compatibility with `restic`, and a repository without that, wouldn't be able to be read by it. Hence, I thought, we keep that, but we make sure, that also without these directories (in case someone deletes them), we are able to create and write to these directories. Not sure, how else we could handle that.

crates/backend/src/local.rs
```diff
@@ -328,26 +348,43 @@ impl WriteBackend for LocalBackend {
...
     /// [`LocalBackendErrorKind::SettingFileLengthFailed`]: LocalBackendErrorKind::SettingFileLengthFailed
     /// [`LocalBackendErrorKind::CouldNotWriteToBuffer`]: LocalBackendErrorKind::CouldNotWriteToBuffer
     /// [`LocalBackendErrorKind::SyncingOfOsMetadataFailed`]: LocalBackendErrorKind::SyncingOfOsMetadataFailed
+    /// [`LocalBackendErrorKind::DirectoryCreationFailed`]: LocalBackendErrorKind::DirectoryCreationFailed
     fn write_bytes(&self, tpe: FileType, id: &Id, _cacheable: bool, buf: Bytes) -> Result<()> {
         trace!("writing tpe: {:?}, id: {}", &tpe, &id);
         let filename = self.path(tpe, id);
+
+        // create parent directory if it does not exist
+        if let Some(parent) = filename.parent() {
```
aawsome, line 369: I would make a method `parent_path` and then just use `self.parent_path(tpe,id)` here.

aawsome (changes requested):
Good catch!
 I have just found a small thing I would change in this PR...

crates/backend/src/local.rs
simonsan: Added!

crates/core/src/commands/check.rs
```diff
@@ -306,7 +306,13 @@ pub(crate) fn check_repository<P: ProgressBars, S: Open>(
 
         packs.into_par_iter().for_each(|pack| {
             let id = pack.id;
-            let data = be.read_full(FileType::Pack, &id).unwrap();
+            let data = match be.read_full(FileType::Pack, &id) {
+                Ok(data) => data,
+                Err(err) => {
+                    error!("Error reading data for pack {id} : {err}");
```
nardoor, line 312: Question: should this error log be different from the one line 318?

nardoor (dismissed):
Hello,

Looks good to me!
Just one question about an  `error` log.

crates/core/src/commands/check.rs
simonsan: Good catch, this is the next step to handle errors in check more gracefully, than currently. So I changed it a bit and added two notes about that.

crates/backend/src/local.rs
aawsome: Actually I thought about having something along the `fn path`, but just without the last `join`. We know there is a parent because we just create paths which do have a parent. My idea was just about removing the error handling which will never be called.

PR #355, opened 2024-11-17
feat(async): add `async_compatible` methods to identify backend compatibility

nardoor:
Add `getters` to warn about `async_incompatible` backends.

Using the type system or fixing all async compatibility issue is big and structural work.
To avoid runtime crash for our users I suggest using this getter as a temporary fix. 

tracking issue: rustic-rs/rustic#1181 (see for more details)

### DIFF
--- a/crates/core/src/backend.rs
+++ b/crates/core/src/backend.rs
@@ -174,6 +174,21 @@ pub trait ReadBackend: Send + Sync + 'static {
     fn warm_up(&self, _tpe: FileType, _id: &Id) -> RusticResult<()> {
         Ok(())
     }
+
+    /// Getter to now if some backend is compatible
+    /// with async features in `rustic_core` implementations.
+    ///
+    /// ## Default impl
+    ///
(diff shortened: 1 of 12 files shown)


nardoor:
For information
`Rclone`, `Rest` and `Opendal` backends are the 3 `async` incompatible backend.

nardoor:
updated the code:
- non api breaking anymore (default impl, async_compatible is opt-in)
- to avoid future double negation, i changed `is_async_incompatible` to `is_async_compatible`

crates/core/src/backend.rs
```diff
@@ -174,6 +174,21 @@ pub trait ReadBackend: Send + Sync + 'static {
     fn warm_up(&self, _tpe: FileType, _id: &Id) -> RusticResult<()> {
         Ok(())
     }
+
+    /// Getter to now if some backend is compatible
```
simonsan, line 178:
```suggestion
    /// Getter to determine if some backend is compatible
```

simonsan (approved):

PR #356, opened 2024-11-17
refactor(backend): migrate from `backoff` to `backon`

nardoor:
Hello,

Removed the unmaintained  `backoff` crate in the `rustic_backend` crate.
Replaced it with the relatively similar `backon` crate.

If `backon` as simpler API, it also lacks a few features that `backoff` had.
For instance: `no_max_delay` or `backoff::retry_notify`. 
To make up for it:
- I opened https://github.com/Xuanwo/backon/pull/160 on the `backon` github.
- I implemented an internal module `backon_extension` inside `backend::rest`.

Let me know what you think of the code!

Fixes https://github.com/rustic-rs/rustic_core/issues/351

### DIFF
--- a/crates/backend/src/rest.rs
+++ b/crates/backend/src/rest.rs
@@ -44,7 +44,7 @@ mod backon_extension {
         fn notify(&self, dur: Duration) {}
     }
 
-    /// A backon::backoff extension that limits the number of retries
+    /// A `backon::backoff` extension that limits the number of retries
     #[derive(Debug)]
     pub struct LimitRetry {
         max_retries: usize,
@@ -58,7 +58,7 @@ mod backon_extension {
         }
     }
 
-    /// We need to impl [`Clone`] manually because [backon::ExponentialBuilder] doesn't.
+    /// We need to impl [`Clone`] manually because [`backon::ExponentialBuilder`] doesn't.
     impl Clone for LimitRetry {
         fn clone(&self) -> Self {
             Self {
@@ -104,10 +104,10 @@ impl backon_extension::NotifyWhenRetry for reqwest::Error {
     ///
     /// Else return `false` and the combined backoff will stop early.
     fn when_retry(&self) -> bool {
-        match self.status() {
-            Some(status_code) => !status_code.is_client_error(), // do no retry if client error
-            None => true,                                        // else retry
-        }
+        self.status().map_or(
+            true,                                         // retry
+            |status_code| !status_code.is_client_error(), // do not retry if `is_client_error`
+        )
     }
 
     /// Notify function for backon in case of error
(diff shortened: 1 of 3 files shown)


aawsome:
Actually I didn't get what the benefit was using `backon_extension` - and I had some problems to understand what it was doing. I tried to simplify things a bit. @nardoor can you have a look at the last commit?

nardoor:
> Actually I didn't get what the benefit was using `backon_extension` - and I had some problems to understand what it was doing. I tried to simplify things a bit. @nardoor can you have a look at the last commit?

Hello,

Thanks for taking a look and contributing.

The extension was replacing what backoff did and backon didn't.

The way I did it could make it easy to reuse the retry in some other backend or other place in the code, or even propose the ideas to backon. 

But If you anticipate we won't need it elsewhere I don't mind moving the logic to `impl Restbackend`

crates/backend/src/rest.rs
```diff
@@ -111,24 +43,27 @@ pub struct RestBackend {
...
 
 impl RestBackend {
+    // call the given operation retrying non-permanent errors and giving warnings for failed operations
+    fn retry_notify<F, T>(&self, op: F) -> Result<T, reqwest::Error>
+    where
+        F: FnMut() -> Result<T, reqwest::Error>,
+    {
+        op.retry(self.backoff)
+            .notify(|err, duration| warn!("Error {err} at {duration:?}, retrying"))
+            .when(|err| {
```
aawsome, line 58: maybe changing the order to `retry(..).when(..).notify(..)` makes it even more clear what is going on here..

nardoor: I agree @aawsome 
I added a commit to do just that.

I also added some doc to the `retry_notify` method.

On a side note, the simplification is indeed nice.
You basically re-did everything for the best. Next time if you don't have time / don't want to bother, I will be pleased to apply analog simplification recommendations.

simonsan (dismissed):

crates/backend/src/rest.rs
```diff
@@ -111,24 +43,37 @@ pub struct RestBackend {
...
+    /// The ``BackoffBuilder`` we use
+    backoff: ExponentialBuilder,
 }
 
 impl RestBackend {
+    /// Call the given operation retrying non-permanent errors and giving warnings for failed operations
+    ///
+    /// ## Permanent/non-permanent errors
+    ///
+    /// - client_error are considered permanent
```
simonsan, line 55:
```suggestion
    /// -  `client_error` are considered permanent
```

simonsan (approved):

aawsome (approved):
LGTM

PR #361, opened 2024-11-18
refactor(error): enhance error logging and output formatting

simonsan:
- adds `display_log()` to `RusticError` for a minimal display to be used in logging
- uses above method in error!/warn!/info! macros over the codebase
- added snapshot test to make sure the output is as expected

### DIFF
--- a/crates/core/src/commands/check.rs
+++ b/crates/core/src/commands/check.rs
@@ -276,7 +276,7 @@ pub(crate) fn check_repository<P: ProgressBars, S: Open>(
         if let Err(err) = cache.remove_not_in_list(FileType::Pack, &ids) {
             warn!(
                 "Error in cache backend removing pack files: {}",
-                err.to_log_output()
+                err.display_log()
             );
         }
         p.finish();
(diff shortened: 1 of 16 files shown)


crates/core/src/commands/check.rs
```diff
@@ -309,21 +312,13 @@ pub(crate) fn check_repository<P: ProgressBars, S: Open>(
...
                     return;
                 }
             };
             match check_pack(be, pack, data, &p) {
                 Ok(()) => {}
-                // FIXME: This needs different handling, now it prints a full display of RusticError
-                // Instead we should actually collect and return a list of errors on the happy path
-                // for `Check`, as this is a non-critical operation and we want to show all errors
-                // to the user.
-                Err(err) => error!("Pack {id} is not valid: {err}",),
```
nardoor, line 326: Are the FIXME not FIXMEs anymore?

I don't know well the `check` cmd so I'm wondering.

crates/core/src/error.rs
```diff
@@ -375,6 +370,54 @@ impl RusticError {
...
+                        format!("Error: {guidance} (kind: {kind}{error_code}): caused by: {cause}")
+                    },
+                    |source| {
+                        format!("Error: {guidance} (kind: {kind}{error_code}): caused by: {cause} : (source: {source})")
+                    },
+                )
+            },
+        );
+
+        output
```
nardoor, line 419: Have you considered using a `fmt: std::fmt::Formatter` in an analog way as `impl Debug` or `impl Display` ?

I suggest that because your 3 branch here are formatted the same, it's just that some are appended.
A formatter could (maybe) simplify the code. (avoiding `map_or_else` and `closures` in favor of `if <cond> { fmt.write!("...")`

nardoor (dismissed):
Hey,

I added a suggestion for the `display_log` impl.
Wether you take it or not the code lgtm!

I see value in having synthetic log for some errors.


---
Just a thought/question:
- if we use display_log more than the `impl Display`:  could we want to invert both implementations ? (`display_full` and `impl Display { fn display(&self) { self.display_log() }  }`

crates/core/src/error.rs
simonsan: Refactored to `write!(fmt , ..)`

PR #365, opened 2024-11-21
feat: add summary module for aggregating runtime information and internal metrics

simonsan:
An informative summary system for aggregating and condensing data collected
from runtime checks, including warnings, issues, and operational metrics.

This system should provide end-users with a clear, concise summary of command
execution results without conflicting with existing error-handling standards.
In scenarios where execution cannot proceed due to a critical error, a
`RusticError` will be raised instead, and no summary will be provided.

# Separation of Concerns

Critical runtime errors that prevent further execution are handled through the
existing `RusticError` system. The `Summary` will only collect information for
non-fatal events.

# Compatibility with Existing Error Handling

Summaries must coexist with error propagation rules. They will not replace
the core behavior of error propagation but act as a complementary mechanism
for presenting non-fatal feedback.

# User-Friendly Reporting

Summaries should aggregate detailed runtime information—such as warnings,
issues, and metrics — in a clear and condensed format for the end-user.

# Aggregation & Condensation

Similar or repeated errors should be aggregated to avoid redundant information,
presenting users with a high-level overview.

### DIFF
--- a/crates/core/src/error/summary.rs
+++ b/crates/core/src/error/summary.rs
@@ -1,0 +1,390 @@
+//! An informative summary system for aggregating and condensing data collected
+//! from runtime checks, including warnings, issues, and operational metrics.
+//!
+//! This system should provide end-users with a clear, concise summary of command
+//! execution results without conflicting with existing error-handling standards.
+//! In scenarios where execution cannot proceed due to a critical error, a
+//! `RusticError` will be raised instead, and no summary will be provided.
+//!
+//! # Separation of Concerns
+//!
+//! Critical runtime errors that prevent further execution are handled through the
+//! existing `RusticError` system. The `Summary` will only collect information for
+//! non-fatal events.
+//!
+//! # Compatibility with Existing Error Handling
+//!
+//! Summaries must coexist with error propagation rules. They will not replace
+//! the core behavior of error propagation but act as a complementary mechanism
+//! for presenting non-fatal feedback.
+//!
+//! # User-Friendly Reporting
+//!
+//! Summaries should aggregate detailed runtime information—such as warnings,
+//! issues, and metrics — in a clear and condensed format for the end-user.
+//!
+//! # Aggregation & Condensation
+//!
+//! Similar or repeated errors should be aggregated to avoid redundant information,
+//! presenting users with a high-level overview.
+
+use std::{
+    collections::{BTreeMap, HashSet},
+    fmt::{self, Display},
+    string::ToString,
+    time::Instant,
+};
+
+pub type IssueIdentifier = String;
+
+pub type Issues = BTreeMap<IssueScope, BTreeMap<IssueIdentifier, CondensedIssue>>;
+pub type Metrics = BTreeMap<String, String>;
+
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    Default,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    derive_more::Display,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub enum IssueScope {
+    #[default]
+    Internal,
+    Unknown,
+    UserInput,
+}
+
+#[derive(Debug, Clone, Default, serde::Serialize, serde::Deserialize)]
+pub struct CondensedIssue {
+    /// High-level description of the problem
+    message: String,
+
+    /// Number of occurrences
+    count: usize,
+
+    /// Optional diagnostic information, e.g. an error message
+    root_cause: Option<String>,
+}
+
+#[derive(Debug, Clone, Copy, Default, Hash, PartialEq, Eq, derive_more::Display)]
+pub enum DisplayOptionKind {
+    #[default]
+    Issues,
+    Timing,
+    Metrics,
+    All,
+}
+
+#[derive(Debug, Clone)]
+pub struct Summary {
+    /// Name of the active context, e.g. a command or operation
+    context: String,
+
+    /// Start time of the collection
+    // Instant cannot be (de-)serialized, for an implementation see:
+    // https://github.com/serde-rs/serde/issues/1375#issuecomment-419688068
+    start_time: Instant,
+
+    /// End time, when the collection is completed
+    // Serialization: See note above
+    end_time: Option<Instant>,
+
+    /// Collection of non-critical warnings   
+    issues: Issues,
+
+    /// Optional custom metrics collected during execution
+    metrics: Metrics,
+
+    /// Display this data
+    display: HashSet<DisplayOptionKind>,
+}
+
+impl Summary {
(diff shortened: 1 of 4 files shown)


crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,415 @@
...
+    Debug, Clone, Copy, Default, Hash, PartialEq, Eq, derive_more::Display, serde::Serialize,
+)]
+pub enum DisplayOptionKind {
+    #[default]
+    Issues,
+    Timing,
+    Metrics,
+    All,
+}
+
```
simonsan, line 85: These could be bitflags or enumflags, but I'm unsure if that wouldn't be overengineering?

simonsan:
What's still missing is functionality to store different categories of issues, e.g. warnings or soft errors. And then to log them to the terminal.

TODO:

- [ ] store different categories of issues (warnings, soft errors)
- [ ] add matching logic to decide which log macro to use
- [ ] add setting to `Summary`, that enables/disables logging on adding issues

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,591 @@
...
+        self.end_time = Some(Instant::now());
+    }
+
+    pub fn enable_log(&mut self) -> Self {
+        self.log_enabled = true;
+        self.clone()
+    }
+
+    pub fn is_error(&self) -> bool {
+        self.issues.contains_key(&IssueCategory::Error)
```
simonsan, line 159: Should probably also check if `self.issues.get(&IssueCategory::Error).len > 0`

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,591 @@
...
+use std::{
+    collections::{BTreeMap, HashSet},
+    fmt::{self, Display, Write},
+    sync::{Arc, Mutex},
+    time::Instant,
+};
+
+use ecow::EcoString;
+
+pub type Issues = BTreeMap<IssueCategory, BTreeMap<IssueScope, CondensedIssue>>;
```
simonsan, line 40: Is `IssueScope` needed, or do we want to deduplicated on some other stuff? We can also not use a `BTreeMap` here, but just a `Vec`tor.

```rust
pub type Issues = BTreeMap<IssueCategory, Vec<CondensedIssue>>;
```

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,592 @@
...
+    // ? How do we merge the context of each summary?
+    // ? We can't merge the context, as it's a unique identifier for each summary.
+    // ? We could add a new field to each CondensedIssue to store the context of the
+    // ? merged summaries.
+    // ?
+    // ? How do we merge the other fields? E.g. Timing, Metrics, etc.
+    pub fn merge(&mut self, other: Self) {
+        self.issues.extend(other.issues);
+        self.metrics.extend(other.metrics);
+    }
```
simonsan, line 296: Use `conflate` and implement `Merge`, make sure we fulfil above requirements and make it as ergonomic as possible. Make sure deduplication works well also for edge cases. We want to be sure, that we can assess if a command soft errored (warnings are lower priority).

simonsan: Ah, conflate is actually behind the `merge` feature flag and doesn't make sense then to be used here.

crates/core/src/error/summary.rs
nardoor: Could `IssueScope` be a member of `CondensedIssue`? 
So you use a `Vec` and still have a notion of `IssueScope`

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,617 @@
...
+    end_time: Option<Instant>,
+
+    /// Collection of non-critical warnings   
+    issues: Issues,
+
+    /// Optional custom metrics collected during execution
+    metrics: Metrics,
+
+    /// Display this data
+    display: HashSet<DisplayOptionKind>,
```
nardoor, line 130: As I'm beginning to read through the code, I am not (yet) fully undertanding what  `DisplayOptionKind` is about.
Could you detail it a bit more in the code-doc? (for instance in the enum's doc)

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,617 @@
...
+
+    /// Marks the summary as completed, capturing the end time.
+    pub fn complete(&mut self) {
+        self.end_time = Some(Instant::now());
+    }
+
+    pub fn enable_log(&mut self) -> Self {
+        self.log_enabled = true;
+        self.clone()
+    }
```
nardoor, line 158: I wonder what is the reason to use clone there?

The builder pattern usually is something like:
```rust
    pub fn enable_log(mut self) -> Self {
        self.log_enabled = true;
        self
    }
```

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,617 @@
...
+                if val.root_cause.is_none() {
+                    val.root_cause.clone_from(&root_cause);
+                }
+            })
+            .or_insert(CondensedIssue {
+                category,
+                message,
+                count: 1,
+                root_cause,
+            });
```
nardoor, line 224: As I see it, when we get in the `and_modify` case, we loose the `message`.
Is that voluntary?

Same we might have != `root_cause` but we'll keep only one.

Wouldn't this lead to "wrong" reporting of some error A that happened N times, but in reality, A happened N1 times and B, C, D, E had the same scope and happened but their "message" and "root_cause" are lost.

I understand we want to aggregate and get a `high level overview`. But as I understand this, we could become `misleading` there.

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// Adds a custom metric
+    pub fn add_metric(&mut self, key: impl Into<EcoString>, value: impl Into<EcoString>) {
+        let value = value.into();
+
+        _ = self
+            .metrics
+            .entry(key.into())
+            .and_modify(|val| *val = value.clone())
+            .or_insert_with(|| value);
+    }
```
nardoor, line 236: Here I see that `metrics` are erase on save.

I feel like this should be documented in the `add` method.

Maybe also this method should be renamed `set_metric`?

crates/core/src/error/summary.rs
```diff
@@ -0,0 +1,617 @@
...
+        self.display.insert(DisplayOptionKind::Timing)
+    }
+
+    pub fn export_metrics(&mut self) -> bool {
+        self.display.insert(DisplayOptionKind::Metrics)
+    }
+
+    pub fn export_all(&mut self) -> bool {
+        self.display.insert(DisplayOptionKind::All)
+    }
```
nardoor, line 252: could you let me know of use cases where we want only some `DisplayOptionKind` and not some others?


Also I feel it's a bit weird that `ALL`  is an enum value.

PR #369, opened 2024-11-25
feat: Improve summary handling and error reporting in repository checks

simonsan:
Enhance the check_repository and check_cache_files functions to return a Summary, improve concurrency handling, and strengthen error logging for better reporting. Refactor summary handling in related functions for consistency.

Based on #365


simonsan:
I'm still debating with myself, if passing in a `summary: Arc<Mutex<Summary>>` into the check functions would be better, instead of returning the `Summary` and initializing it at the beginning of the functions, calling `complete` at the end. For now, we have clearer contexts and timings due to that. But we could also add that functionality based on passing in a summary. We could impl `start/stop` on `Summary`, for timing purposes.

I do think it's a bit cleaner separation, to return a `Summary` and then merge it with the `Summary` up in the call stack.

PR #374, opened 2024-11-28
feat: Add a "minutely" timeline

jullanggit:
## Disclaimer
I am not very experienced with contributing to projects like rustic, so please correct me if I make any mistakes or could improve something about this PR!
In this case, I am particularly unsure about the commit message and how to add tests, so any tips there would be appreciated :smile:
## Motivation
As creating a rustic snapshot can be really fast if not much has changed (less than 10s), I have started to backup-up my files every ten minutes, but noticed that rustic doesn't yet have any keep-options for sub-hourly snapshots, so I decided to add them!

### DIFF
--- a/crates/core/src/commands/forget.rs
+++ b/crates/core/src/commands/forget.rs
@@ -426,6 +426,7 @@ impl KeepOptions {
     /// # Returns
     ///
     /// The list of reasons why the snapshot should be kept
+    #[allow(clippy::too_many_lines)]
     fn matches(
         &mut self,
         sn: &SnapshotFile,
--- a/crates/core/src/commands/forget.rs
+++ b/crates/core/src/commands/forget.rs
@@ -379,7 +379,7 @@ fn equal_hour(sn1: &SnapshotFile, sn2: &SnapshotFile) -> bool {
 ///
 /// # Returns
 ///
-/// Whether the hours of the snapshots are equal
+/// Whether the minutes of the snapshots are equal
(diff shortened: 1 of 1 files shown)


crates/core/src/commands/forget.rs
```diff
@@ -356,12 +370,31 @@ fn equal_hour(sn1: &SnapshotFile, sn2: &SnapshotFile) -> bool {
...
+/// Evaluate the minutes of the given snapshots
+///
+/// # Arguments
+///
+/// * `sn1` - The first snapshot
+/// * `sn2` - The second snapshot
+///
+/// # Returns
+///
+/// Whether the hours of the snapshots are equal
```
aawsome, line 382: should be `minutes` here..

simonsan:
Please also add an explanation to `./config/full.toml` and the `./config/Readme.md` 📓

aawsome:
@jullanggit Ah, I think these are due to a new Rust version. We should fix those in a separate PR before we merge this one...

aawsome:
It would be great if you could extend/adapt the unittests in `forget.rs` to cover the new option, @jullanggit Besides this, this PR looks good to me, just needs to wait until we fix the clippy lints in master. Thanks a lot for the contribution!

simonsan:
The only thing missing now for clippy is the `#[allow(clippy::too_many_lines)]` annotation on `matches()`.

simonsan (approved):

simonsan:
Thank you for the contribution! 🚀 :)

PR #378, opened 2024-12-02
fix(chunker): Don't underflow with wrong size_hint

aawsome:
see #377

### DIFF
--- a/crates/core/src/chunker.rs
+++ b/crates/core/src/chunker.rs
@@ -16,8 +16,8 @@ pub(super) mod constants {
     pub(super) const MIN_SIZE: usize = 512 * KB;
     /// The maximum size of a chunk.
     pub(super) const MAX_SIZE: usize = 8 * MB;
-    /// Buffer size used for reading.
-    pub(super) const BUF_SIZE: usize = 64 * KB;
+    /// Buffer size used for reading - TODO: Find out optimal size for best performance!
+    pub(super) const BUF_SIZE: usize = 4 * KB;
     /// Random polynomial maximum tries.
     pub(super) const RAND_POLY_MAX_TRIES: i32 = 1_000_000;
 }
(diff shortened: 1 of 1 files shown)


crates/core/src/chunker.rs
```diff
@@ -16,8 +16,8 @@ pub(super) mod constants {
     pub(super) const MIN_SIZE: usize = 512 * KB;
     /// The maximum size of a chunk.
     pub(super) const MAX_SIZE: usize = 8 * MB;
-    /// Buffer size used for reading.
-    pub(super) const BUF_SIZE: usize = 64 * KB;
+    /// Buffer size used for reading - TODO: Find out optimal size for best performance!
+    pub(super) const BUF_SIZE: usize = 4 * KB;
```
nardoor, line 20: common kernel PAGE_SIZE would also be my **guess** ^^

crates/core/src/chunker.rs
```diff
@@ -165,7 +163,7 @@ impl<R: Read + Send> Iterator for ChunkIter<R> {
             self.pos += 1;
             self.rabin.slide(byte);
         }
-        self.size_hint -= vec.len();
+        self.size_hint = self.size_hint.saturating_sub(vec.len()); // size_hint can be too small!
```
nardoor, line 166: nice
do we want to also give an example of that happening here ? (like linux pseudo fs in /proc)

nardoor (approved):
:+1:

crates/core/src/chunker.rs
aawsome: I think the main use case is just that between `stat`ing a file and opening it, there could be another process which has modified the file - potentially making it larger..
But I would not add an explicit example; it is simply preventing an underflow here which just makes sense...

crates/core/src/chunker.rs
aawsome: Yes, but we don't have performance tests yet. Maybe there is a better one.
Also, we could find out the performance results differ on different systems/plattforms. Then we might want to expose this as a parameter to users..

PR #379, opened 2024-12-04
feat(warmup)!: Add warmup wait command

tinou98:
This allows to programmatically (with script) wait for elements to be ready, instead of hardcoding a duration.

<details><summary>Scripts examples</summary>

Warm-up:
```bash
KEY="data/${1:0:2}/$1";
if [ "$(aws s3api head-object --bucket cold --key "$KEY" --query StorageClass)" == '"GLACIER"' ]
then
    aws s3api restore-object --restore-request Days=7 --bucket cold --key "$KEY"
fi
```  

Warm-up wait:
```bash
KEY="data/${1:0:2}/$1";
until [ "$(aws s3api head-object --bucket cold --key "$KEY" --query StorageClass)" != '"GLACIER"' ]
do
    sleep 60s
done
```
</details>

This is a simple version that will wait for each object sequentially (but that shouldn't be a problem as all warm up request are done in a first pass, and thus runs in parallel). 

I'm up for discussion if another model looks good, like
- moving the loop in the lib, and the `wait` script just check if the file is ready
- having only one script, that warm-up and then return the status

### DIFF
--- a/crates/core/src/repository.rs
+++ b/crates/core/src/repository.rs
@@ -157,6 +157,14 @@ pub struct RepositoryOptions {
     #[cfg_attr(feature = "merge", merge(strategy = conflate::option::overwrite_none))]
     pub warm_up_command: Option<CommandInput>,
 
+    /// Wait for end of warm up by running the command with %id replaced by pack id
+    #[cfg_attr(
+        feature = "clap",
+        clap(long, global = true, conflicts_with = "warm_up_wait",)
+    )]
+    #[cfg_attr(feature = "merge", merge(strategy = conflate::option::overwrite_none))]
+    pub warm_up_wait_command: Option<CommandInput>,
+
     /// Duration (e.g. 10m) to wait after warm up
     #[cfg_attr(feature = "clap", clap(long, global = true, value_name = "DURATION"))]
     #[serde_as(as = "Option<DisplayFromStr>")]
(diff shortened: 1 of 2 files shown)


crates/core/src/repository.rs
```diff
@@ -748,7 +756,10 @@ impl<P: ProgressBars, S> Repository<P, S> {
     ///
     /// * If the command could not be parsed.
     /// * If the thread pool could not be created.
-    pub fn warm_up_wait(&self, packs: impl ExactSizeIterator<Item = PackId>) -> RusticResult<()> {
+    pub fn warm_up_wait(
+        &self,
+        packs: impl ExactSizeIterator<Item = PackId> + Clone,
```
simonsan, line 761: Why do we require `Clone` here? This would be an API breaking change 🤔

tinou98: We need to iterate twice on the list, the first time to warm up, the second time to wait on each element.

If we want to avoid this change, we can collect the list, but that would introduce an allocation.

I can change to a collect and remove the `Clone` requirement if we want to keep the API stable.

aawsome: Actually I would prefer to leave it and merge this PR with the next breaking release. I doubt that the current public API of `warm_up` is externally used anyway..

simonsan: Yeah, sure. It feels a bit like premature optimization to me, though, not sure if it's needed at this point. If it's not really needed and there is a way around (just allocate) I don't think it is necessary to introduce a bound here at this stage - if we need to get rid of it, for any reason, this is another API breaking change. So I would be vary now.

aawsome: I propose to change the visibility to `pub(crate)` - this is a breaking change but then future changes won't be breaking anymore.

simonsan: > I propose to change the visibility to `pub(crate)` - this is a breaking change but then future changes won't be breaking anymore.

That would be also a solution, we need to check, if that would also change the Result type then to a local error. I didn't look into that, can do later.

simonsan: Looked through `warm_up_wait`, we **wouldn't** need to change the Return value to a local one. So we can just make it `pub(crate)` as proposed.

tinou98: I updated the warm_up_wait command to be crate local (and rebased)

PR #381, opened 2024-12-06
fix!: Don't panic when reading empty files

aawsome:
There was a bug when using partition_point which in that case does not gave guarentees about its behavior. Hence the whole logic was refactored and unit/integration tests were added.

closes https://github.com/rustic-rs/rustic/issues/1380

### DIFF
--- a/crates/core/src/vfs.rs
+++ b/crates/core/src/vfs.rs
@@ -456,7 +456,7 @@ impl Vfs {
 pub struct OpenFile {
     // The list of blobs
     content: Vec<DataId>,
-    offsets: ContentOffset,
+    startpoints: ContentStartpoints,
 }
 
 impl OpenFile {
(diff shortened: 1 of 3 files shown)


crates/core/src/vfs.rs
```diff
@@ -488,26 +479,15 @@ impl OpenFile {
...
-            .as_ref()
-            .unwrap()
-            .iter()
-            .map(|id| {
-                let starts_at = start;
-                start += repo.index().get_data(id).unwrap().data_length() as usize;
-                BlobInfo { id: *id, starts_at }
-            })
-            .collect();
+        let content: Vec<_> = node.content.as_ref().unwrap().clone();
```
simonsan, line 482: Can we replace the `unwrap` with an `expect` here, as we aim to reduce the number of unwraps (without any further information), I thought. We should probably return a `RusticResult<Self>` here though and use an error, because `rustic_core` shouldn't panic as a library as that's harder to catch on the consumer side.

crates/core/src/vfs.rs
```diff
@@ -488,26 +479,15 @@ impl OpenFile {
...
 
-        // content is assumed to be partitioned, so we add a starts_at:MAX entry
-        content.push(BlobInfo {
-            id: DataId::default(),
-            starts_at: usize::MAX,
-        });
+        let offsets = ContentOffset::from_sizes(
+            content
+                .iter()
+                .map(|id| repo.index().get_data(id).unwrap().data_length() as usize),
```
simonsan, line 487: Same here, let's not unwrap in any refactored, new code.

crates/core/src/vfs.rs
```diff
@@ -530,19 +510,16 @@ impl OpenFile {
...
-        mut offset: usize,
+        offset: usize,
         mut length: usize,
     ) -> RusticResult<Bytes> {
-        // find the start of relevant blobs => find the largest index such that self.content[i].starts_at <= offset, but
-        // self.content[i+1] > offset  (note that a last dummy element has been added)
-        let mut i = self.content.partition_point(|c| c.starts_at <= offset) - 1;
-
-        offset -= self.content[i].starts_at;
+        let (mut i, mut offset) = self.offsets.offset(offset);
```
simonsan, line 516: `let (mut i, mut offset) = self.offsets.offset(offset);` that looks like a code smell imho, as we are using four times the same term in a single line.

Maybe we can make variables or methods/functions more self-descriptive here.

crates/core/src/vfs.rs
simonsan: It's fine to make this api breaking, as we have a lot of other api breaking code already coming up.

aawsome: here we can actually just use an empty vec.

crates/core/src/vfs.rs
aawsome: But yes, here we need error handling. I made this `pub(crate)` as the preferred way is to use `Repository::open_file` - which already returns `Result`.
Is a breaking change now.

crates/core/src/vfs.rs
aawsome: done.

simonsan (approved):

PR #383, opened 2024-12-10
refactor!: Add more control over used keys

aawsome:
Adds more control over the used repository keys.
This is needed to implement the missing `key list`, `key remove` commands in the rustic CLI.

As this refactors the `Repository` a bit, this is a breaking change (which will most likely not harm anyone using rustic_core)

### DIFF
--- a/crates/core/src/repofile/keyfile.rs
+++ b/crates/core/src/repofile/keyfile.rs
@@ -9,6 +9,7 @@ use crate::{
     crypto::{aespoly1305::Key, CryptoKey},
     error::{ErrorKind, RusticError, RusticResult},
     impl_repoid,
+    repofile::RepoFile,
 };
 
 /// [`KeyFileErrorKind`] describes the errors that can be returned for `KeyFile`s
@@ -73,6 +74,12 @@ pub struct KeyFile {
     /// The salt used with `scrypt`
     #[serde_as(as = "Base64")]
     salt: Vec<u8>,
+}
+
+impl RepoFile for KeyFile {
+    const TYPE: FileType = FileType::Key;
+    const ENCRYPTED: bool = false;
+    type Id = KeyId;
 }
 
 impl KeyFile {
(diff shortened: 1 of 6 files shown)


crates/core/src/repository.rs
```diff
@@ -863,12 +847,33 @@ impl<P, S: Open> Repository<P, S> {
...
+    pub fn key_id(&self) -> &KeyId {
+        &self.open_status().key_id
+    }
+
+    /// Delete the key with id starting with the given string from the repository.
+    ///
+    /// # Errors
+    ///
+    /// * If the key could not be removed.
+    pub fn delete_key(&self, id: &str) -> RusticResult<()> {
```
simonsan, line 868: Two things:

- what happens in case two keys have the same starting strings?
- there is no safe-guarding in place, before deleting the keys (although that might fit better into `rustic-rs`

aawsome: `find_id` returns an error if no unique id can be found.
The safeguard is the check against `key_id()` - which is the key used to open the repo.

aawsome:
rebased this PR to current master

aawsome:
@simonsan is there anything open which prevents this PR from being merged? (I think we should anyway release a 0.8.0 version, so IMO the breaking changes are no blockers...)

crates/core/src/repofile/keyfile.rs
```diff
@@ -45,34 +46,40 @@ impl_repoid!(KeyId, FileType::Key);
...
 
     /// The key data encrypted by `scrypt`
     #[serde_as(as = "Base64")]
-    data: Vec<u8>,
+    pub data: Vec<u8>,
 
     /// The salt used with `scrypt`
     #[serde_as(as = "Base64")]
-    salt: Vec<u8>,
+    pub salt: Vec<u8>,
```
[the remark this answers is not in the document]
aawsome: This is like we already do it for all other repository files which are anyways part of the repository format specification. 

About API: we have serde::Serialize in the API which allows every user to serialize to a JSON. So I don't think we over-exposure here;-)

simonsan (approved):

PR #399, opened 2025-02-21
chore: update to 2024 edition and fix clippy lints

simonsan:
## Needs

- [x] cargo-deny-action needs to support 2024 edition/rust 1.85
- [x] cargo tarpaulin needs to support 2024 edition/rust 1.85


aawsome (approved):
LGTM

PR #409, opened 2025-05-31
refactor: separate backend operations

### LINKED ISSUE #146: The `copy` command may copy duplicate data
The `copy` command may copy duplicated blob.

How to reproduce:

1. Prepare a file that is large enough, for example, `ffmpeg.exe`
2. Copy `ffmpeg.exe` to `ffmpeg2.exe`, `ffmpeg3.exe`
3. Create a config file like this:
``` toml
[repository]
repository="source"
password="123456"

[[copy.targets]]
repository="target"
password="789012"
```
4. Save the config file as `config.toml`.
5. Create some snapshots:
``` bash
rustic -P config backup ffmpeg.exe
rustic -P config backup ffmpeg2.exe
rustic -P config backup ffmpeg3.exe
```
6. Copy source to target: `rustic -P config copy`
7. Compare sizes of two repositories and the target repository will be larger.
8. Prune the target repository and it will reported that there are unused data.

It seems that when copying data blobs in `src/command/copy.rs`, the index is not updated so duplicate data blobs are copied.

> *I realized this when copying a 300GB repo into a 512GB hard disk and encountered a surprising disk full.*

aawsome:
This PR refactors indexer and packer into a version which is completely independent from I/O and concurrency but simply provides the functionality.
This can then be separated into real "core" functionality.

As a side effect, this PR solves the following topics:
- Ensure that really no duplicate blobs are saved (closes #146)
- Solves potential race conditions (which in theory could lead to data loss)  between finalizing packer/indexed and file_saver by using `self` in the `finalize` methods.


crates/core/src/blob/packer.rs
```diff
@@ -511,16 +148,8 @@ impl<BE: DecryptWriteBackend> RawPacker<BE> {
...
-            err.overwrite_kind(ErrorKind::Internal)
-                .prepend_guidance_line("Failed to save packfile. Data may be lost.")
-                .ask_report()
-        })?;
-
-        self.file_writer.take().unwrap().finalize()?;
-
-        Ok(std::mem::take(&mut self.stats))
+    pub fn finalize(mut self) -> RusticResult<(Option<(Bytes, IndexPack)>, PackerStats)> {
+        Ok((self.save()?, self.stats))
```
simonsan, line 152: Not sure about the error handling here, we would not have an insight here, if we bubble it up just like that, before it was wrapped.

crates/core/src/index/indexer.rs
```diff
@@ -71,38 +61,46 @@ impl<BE: DecryptWriteBackend> Indexer<BE> {
...
     /// * If the index file could not be serialized.
-    pub fn finalize(&self) -> RusticResult<()> {
+    pub fn finalize(mut self) -> Option<IndexFile> {
+        self.save()
+    }
+
+    pub fn needs_save(&self) -> bool {
+        // check if IndexFile needs to be saved
+        let elapsed = self.created.elapsed().unwrap_or_else(|err| {
+            warn!("couldn't get elapsed time from system time: {err:?}");
```
simonsan, line 94: This should be part of the error handling here and not a warning in the logs.

crates/core/src/blob/repopacker.rs
```diff
@@ -0,0 +1,474 @@
...
+        configfile::ConfigFile,
+        indexfile::{IndexBlob, IndexPack},
+        packfile::PackId,
+    },
+};
+
+use super::{
+    pack_sizer::{DefaultPackSizer, FixedPackSizer, PackSizer},
+    packer::{Packer, PackerStats},
+};
```
simonsan, line 27: Imports not from the root

crates/core/src/blob/repopacker.rs
```diff
@@ -0,0 +1,474 @@
...
+        let (tx_raw, rx_raw) = bounded(0);
+        let (finish_tx, finish_rx) = bounded::<RusticResult<PackerStats>>(0);
+        let repository_packer = Self {
+            indexer: indexer.clone(),
+            sender: tx,
+            raw_sender: tx_raw.clone(),
+            finish: finish_rx,
+        };
+
+        let _join_handle = std::thread::spawn(move || {
```
simonsan, line 112: Don't we want to use the join handles somewhere and join the threads, e.g. in case we get Ctrl + C or something?

crates/core/src/blob/repopacker.rs
```diff
@@ -0,0 +1,474 @@
...
+        drop(self.sender);
+        drop(self.raw_sender);
+        // wait for items in channels to be processed
+        self.finish
+            .recv()
+            .expect("Should be able to receive from channel to finalize packer.")
+    }
+}
+
+// TODO: add documentation
```
simonsan, line 236: Docs

crates/core/src/blob/repopacker.rs
```diff
@@ -0,0 +1,474 @@
...
+        Ok(index)
+    }
+
+    fn index(&self, index: IndexPack) -> RusticResult<()> {
+        self.indexer
+            .add_and_check_save(index, false, |file| self.be.save_file_no_id(file))
+    }
+}
+
+// TODO: add documentation
```
simonsan, line 267: Docs

simonsan (commented):
Love the direction, feel like there is a bit more error handling needed at certain places. Need to take another, deeper look at `RepoPacker`, few TODOs and docs still open.

crates/core/src/blob/packer.rs
aawsome: Actually, the errors can only occur when writing the pack header in `write_header`. And those are all very well handled errors, IMO. I don't see a reason why we should handle them in another way in `finalize` than we're handling them in `save`...

crates/core/src/index/indexer.rs
aawsome: This is the error handling. When we can't get an elapsed time when we check if an index file should be saved, we have two options:
1) return an error which would mean we don't finish the backup
2) omit the check on the elapsed time, i.e. only write when the index file is "full" and warn the user.  This will successfully finish the backup.

I strongly prefer option 2) which is the current error handling.

crates/core/src/blob/repopacker.rs
aawsome: added

crates/core/src/blob/repopacker.rs
aawsome: added

crates/core/src/index/indexer.rs
simonsan: This wasn't what I meant, what I meant was if we have a warning in the log, user of rustic_core are not able to handle this, instead of a log warning this should return a rustic_error with the message so a user can handle this if we get to it from somewhere.

nardoor (commented):
I think these are nice API changes, it's good to see the `BE` backend generic type be less propagated if that makes sense.

PR #412, opened 2025-06-26
fix: improve handling of u32 conversions

aawsome:
As packs are only allowed up to a size of 4GiB, some internal data structures use `u32` instead of `usize` or `u64` to save space. This PR improves the handling of errors in the needed type conversations. As we assume at least a 32bit systems, conversions from `u32` to `usize` will panic.


crates/core/src/commands/restore.rs
```diff
@@ -743,15 +743,9 @@ impl RestorePlan {
...
-            let usize_length = usize::try_from(length).map_err(|err| {
-                RusticError::with_source(
-                    ErrorKind::Internal,
-                    "Failed to convert the length `{length}`  to usize. Please try again.",
-                    err,
-                )
-                .attach_context("length", length.to_string())
-            })?;
+            let usize_length =
+                usize::try_from(length).expect("convert from u32 to usize should not fail!");
```
simonsan, line 747: I think we shouldn't introduce new panics here, and instead just throw an error if we can. So users of rustic_core can handle this. An expect is final. We should improve the erroe message maybe.

aawsome: This will only panic on machine architectures where usize has less than 32 bit. For this architectures, we have much more problems, e.g. that we cannot access enough memory to keep the pack (which can be up to 4GiB in size) in-memory.  In fact less than 32 bit in reality means that we have a 16 bit usize meaning we can address 64k of memory - that is nothing rustic is ever meant for.

Anyway, I just realized that this part in the code can be completely eleminated by allowing to hash from a reader. In this case we don't need to convert to `usize`, but use an `u64` to limit the reader. As a side effect, we no longer need to allocate the whole blob in memory which improves performance and memory allocations. So I just changed this.

PR #413, opened 2025-06-28
Add padding blob to data packs to mitigate chunking attacks

aawsome:
see e.g. https://github.com/rustic-rs/rustic/discussions/1439
This also introduces the repository config option `use-pack-padding` which allows to disable the padding.

As a side-effect currently wrong statistics (data added to blobs in stats was without pack header) has been corrected.

depends on #409


crates/core/src/blob/packer.rs
```diff
@@ -346,4 +344,64 @@ impl<C: CryptoKey, S: PackSizer> Packer<C, S> {
...
+
+fn padding_size(size: u32) -> u32 {
+    // compute padding size. Note that we don't add zero-sized blobs here, i.e. padding_size is in 1..=MAX_PADDING.
+    let padding = constants::MAX_PADDING - size % constants::MAX_PADDING;
+    if padding == 0 {
+        constants::MAX_PADDING
+    } else {
+        padding
+    }
+}
```
nardoor, line 387: I took a look at the restic PR for padding attack resilience, it seems they implemented the "padmé" padding "size" algorithm for this.

They reference this blogpost https://lbarman.ch/blog/padme/ and it seems interesting (you probably already have seen it since you participated in the restic issue on chunking attacks).

What's your opinion on that?  
I am not strongly opinionated on this, but it seems that padmé is a thought through padding algorithm with a good balance between security and overhead.

Issue #398, opened 2025-02-09
Add a method to address snapshots relative to the latest / general relative snapshot adressing


Cookie04DE:
It would be very useful to be able to address snapshots not just with `latest` but also with something like `latest-1`, for example when diffing a file to figure out what happened between the current and previous snapshot.

Sort of like how to address git commits with `HEAD` and `HEAD~1`.

In a similar vain something like this with ids would be useful too, so one can address a snapshot and it's parent snapshots.

Maybe even in the positive ("future") direction, but I am not familiar enough with the code to know if that is feasible.

nardoor:
Hello @Cookie04DE 

Thank you for taking the time to posting a feature request.
I am interested in doing the work to implement that, updates will be posted on this issue.

Thanks again and have a nice weekend.

PR #416, opened 2025-07-13
feat: snapshots can be identified with latest~N (N >= 0)

nardoor:
This is a feature request from #398

### DIFF
--- a/crates/core/src/repofile/snapshotfile.rs
+++ b/crates/core/src/repofile/snapshotfile.rs
@@ -525,6 +525,8 @@ impl SnapshotFile {
 
     /// Get a [`SnapshotFile`] from the backend by (part of the) Id
     ///
+    /// Works with a snapshot `Id` or a `latest` indexed syntax: `latest` or `latest~N` with N >= 0
+    ///
     /// # Arguments
     ///
     /// * `be` - The backend to use
@@ -537,16 +539,38 @@ impl SnapshotFile {
     /// * If the string is not a valid hexadecimal string
     /// * If no id could be found.
     /// * If the id is not unique.
+    /// * If the `latest` syntax is "detected" but inexact
     pub(crate) fn from_str<B: DecryptReadBackend>(
         be: &B,
         string: &str,
         predicate: impl FnMut(&Self) -> bool + Send + Sync,
         p: &impl Progress,
     ) -> RusticResult<Self> {
-        match string {
-            "latest" => Self::latest(be, predicate, p),
-            _ => Self::from_id(be, string),
+        if string == "latest" {
+            return Self::latest(be, predicate, p);
         }
+        let err = RusticError::new(
+                    ErrorKind::InvalidInput,
+                    "Invalid snapshot identifier \"{input}\". Expected either a \"snapshot_id\" or \"latest\" or \"latest~N\" (N >= 0).",
+                )
+                .attach_context("input", string);
+
+        // we do a broad match to give a contextual error in case
+        // the user tried to use the latest~N feature
+        if string.contains('~') || string.contains("latest") {
+            let Some((latest_prefix, n)) = string.split_once('~') else {
+                return Err(err);
+            };
+            if latest_prefix != "latest" {
+                return Err(err);
+            }
+            let Ok(n) = n.parse::<usize>() else {
+                return Err(err);
+            };
+            return Self::latest_n(be, predicate, p, n);
+        }
+
+        Self::from_id(be, string)
     }
 
     /// Get the latest [`SnapshotFile`] from the backend
@@ -565,10 +589,35 @@ impl SnapshotFile {
         predicate: impl FnMut(&Self) -> bool + Send + Sync,
         p: &impl Progress,
     ) -> RusticResult<Self> {
-        p.set_title("getting latest snapshot...");
+        Self::latest_n(be, predicate, p, 0)
+    }
+
+    /// Get the latest [`SnapshotFile`] from the backend
+    ///
+    /// # Arguments
+    ///
+    /// * `be` - The backend to use
+    /// * `predicate` - A predicate to filter the snapshots
+    /// * `p` - A progress bar to use
+    /// * `n` - The n-latest index to go back for snapshot
+    ///
+    /// # Errors
+    ///
+    /// * If no snapshots are found
+    pub(crate) fn latest_n<B: DecryptReadBackend>(
+        be: &B,
+        predicate: impl FnMut(&Self) -> bool + Send + Sync,
+        p: &impl Progress,
+        n: usize,
+    ) -> RusticResult<Self> {
+        if n == 0 {
+            p.set_title("getting latest snapshot...");
+        } else {
+            p.set_title("getting latest~N snapshot...");
+        }
         let mut latest: Option<Self> = None;
         let mut pred = predicate;
-
+        let mut snapshots = Vec::new();
         for snap in be.stream_all::<Self>(p)? {
             let (id, mut snap) = snap?;
             if !pred(&snap) {
@@ -576,12 +625,13 @@ impl SnapshotFile {
             }
 
             snap.id = id;
-            match &latest {
-                Some(l) if l.time > snap.time => {}
-                _ => {
-                    latest = Some(snap);
-                }
-            }
+            snapshots.push(snap);
+        }
+
+        // sort in decreasing time order
+        snapshots.sort_by(|s1, s2| s2.time.cmp(&s1.time));
+        if snapshots.len() > n {
+            latest = Some(snapshots.swap_remove(n));
         }
 
         p.finish();
--- a/crates/core/src/repofile/snapshotfile.rs
+++ b/crates/core/src/repofile/snapshotfile.rs
@@ -23,6 +23,7 @@ use crate::{
     backend::{FileType, FindInBackend, decrypt::DecryptReadBackend},
     blob::tree::TreeId,
     error::{ErrorKind, RusticError, RusticResult},
+    id::constants::HEX_LEN,
     impl_repofile,
     progress::Progress,
     repofile::RepoFile,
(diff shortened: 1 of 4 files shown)


crates/core/src/repofile/snapshotfile.rs
```diff
@@ -537,16 +539,38 @@ impl SnapshotFile {
...
+        }
+        let err = RusticError::new(
+                    ErrorKind::InvalidInput,
+                    "Invalid snapshot identifier \"{input}\". Expected either a snapshot id: \"01a2b3c4\" or \"latest\" or \"latest~N\" (N >= 0).",
+                )
+                .attach_context("input", string);
+
+        // we do a broad match to give a contextual error in case
+        // the user tried to use the latest~N feature
+        if string.contains('~') || string.contains("latest") {
```
aawsome, line 560: I would work with `strip_prefix("latest~")` here. We can either have that prefix and then treat the `N` or we can normally parse the id. This also removes a bit of the error handing below

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -1441,4 +1509,166 @@ mod tests {
...
+                ),
+                (
+                    // this is the latest
+                    id3,
+                    DateTime::from_timestamp(1_752_483_800, 0).unwrap().into(),
+                ),
+            ],
+            &key,
+        );
+
```
aawsome, line 1623: this is duplicate to the above code. Can you make a single set-up func for this?

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -1441,4 +1509,166 @@ mod tests {
...
+        let mut back = MockBackend::new();
```
aawsome, line 1562: While using the `MockBackend` is a nice example how to use this Mock, we can actually use the `InMemoryBackend` here as well for these unit tests, see the integration tests under `tests`. I think this could be even a bit easier and also uses a HashMap in the background.

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -565,32 +589,65 @@ impl SnapshotFile {
...
             }
 
             snap.id = id;
-            match &latest {
-                Some(l) if l.time > snap.time => {}
-                _ => {
-                    latest = Some(snap);
-                }
-            }
+            snapshots.push(snap);
```
aawsome, line 628: Here I think using something like `.filter()`/`.map()` `.collect()` would be more idiomatic Rust.
Also as a side note: Collecting all snapshots removes the (small) memory optimization of not storing all snapshots in-memory. But I think we can neglect this and simplify the code - especially with the goal in mind to separate backend operations from the snapshot selection.

crates/core/src/repofile/snapshotfile.rs
aawsome: Actually we could work with a two-step `strip_prefix`: First try to strip `latest`. If not possible, parse the id. If possible, check for the remainder being empty (and if return the `latest` snapshot). Else try to strip the prefix `~` and treat the remainder as number `N` or error if not able to strip.

crates/core/src/repofile/snapshotfile.rs
nardoor: Oh I definitely could have used that. I will certainly use that in future utests.

I could leave it like that for `mockall` references what do you think about that?

crates/core/src/repofile/snapshotfile.rs
nardoor: I updated it :)

crates/core/src/repofile/snapshotfile.rs
nardoor: I saw some value in trying to detect a "wrong" usage of the latest~N syntax (typo in latest OR wrong separator sign)

But I can see how it's more important to simplify the code :), applied!

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -617,19 +614,19 @@ impl SnapshotFile {
...
-                continue;
-            }
-
-            snap.id = id;
-            snapshots.push(snap);
-        }
+        let mut snapshots: Vec<_> = be
+            .stream_all::<Self>(p)?
+            .into_iter()
+            .filter_map(Result::ok)
```
aawsome, line 619: Note that this chenges the error handling: Now all errors when reading a single SnapshotFile from the backend are ignored. This is somehow desired behavior, IMO, but deserves a warning. I added one ;-)

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -565,32 +586,65 @@ impl SnapshotFile {
...
+            .filter(|(_id, snap)| pred(snap))
+            .map(|(id, mut snap)| {
+                snap.id = id;
+                snap
+            })
+            // sort in decreasing time order
+            .sorted_by(|s1, s2| s2.time.cmp(&s1.time))
```
aawsome: `Itertools::sorted_by` does collect the iterator into a `Vec`, sort it (note there is `Itertools::sorted_unstable_by` which is more performant) and provide another iterator over the sorted `Vec`. I believe that the compiler will be able to optimize the formally extra `.into_iter().collect()` away we get here, but I still believe that in this case collecting the original iterator and sort the `Vec` by yourself would be more straightforward and clear without adding code complexity.

However, when looking into `Itertools` there is also `k_smallest_by` which does not collect the iterator but only allocates a binary heap with the size of `N` - which is IMO the optimal solution for the problem here. Plus, they tell that collecting the resulting iterator will not allocate, but just return the heap as `Vec`. So I changed this here to use this nice functionality.

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -565,32 +586,65 @@ impl SnapshotFile {
...
+            p.set_title("getting latest~N snapshot...");
+        }
         let mut latest: Option<Self> = None;
         let mut pred = predicate;
+        let mut snapshots: Vec<_> = be
+            .stream_all::<Self>(p)?
+            .into_iter()
+            .filter_map(Result::ok)
+            .filter(|(_id, snap)| pred(snap))
+            .map(|(id, mut snap)| {
```
aawsome: In other methods, we use `set_id`. So, I changed this here.

aawsome (commented):
I commited some small optimazations.

There is however one thing we should take into account: The `Snapshotfile::from_str` method is not used by all rustic commands. Especially `rustic diff` uses another method only accepting (multiple) ids which will not be affected by this change. Also `rustic snapshots latest` uses a method which groups snapshots and looks for the latest of each group - which is also not affected by this change.

So, to get a complete support we need to add this to other methods - why I strongly recommend we refactor this (and generally the snapshots handling) into smaller pieces which can be reused..

nardoor:
Hello,


I still have in mind the refactor we talked about. I know how I'd do it, just have to take a moment to propose it.

Also I agree that snapshot identification for the user should be normalized for all commands!
I'll take a look at the different commands in a PR in rustic after this one is done.

aawsome:
> Hello,
> 
> I still have in mind the refactor we talked about. I know how I'd do it, just have to take a moment to propose it.
> 
> Also I agree that snapshot identification for the user should be normalized for all commands! I'll take a look at the different commands in a PR in rustic after this one is done.

Please note that it is not that easy: for some commands like `rustic snapshots` or `rustic tag` etc. it might be desirable to specify multiple snapshots. And it would be *very* inefficient to load them by looping over the current `from_str` method - for something like `latest*` this involves loading and parsing all snapshots! 

So, yes, we need to take a look at the `rustic` commands and uniformly treat snapshot parsing+loading. But we also need to implement the needed algorithms to efficiently support a request of multiple snapshots - including something like `latest abc123 <FULL_HEX> latest~5` - which we currently don't have.

aawsome:
@nardoor can you please review the commits I added? Thanks!

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -546,28 +586,11 @@ impl SnapshotFile {
...
-        let Some(suffix) = string.strip_prefix("latest") else {
-            return Self::from_id(be, string);
-        };
-
-        if suffix.is_empty() {
-            return Self::latest(be, predicate, p);
+        match string.parse()? {
+            SnapshotRequest::Latest(n) => Self::latest_n(be, predicate, p, n),
+            SnapshotRequest::StartsWith(id) => Self::from_id(be, &id),
+            SnapshotRequest::Id(id) => Self::from_backend(be, &id),
```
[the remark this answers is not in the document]
aawsome: > Should we anticipate with some APIs that take a Vec of SnapshotRequest and resolves them in an efficient way?

exactly that was what I had in mind! But yes, let's postpone this to a future PR.

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -853,16 +869,18 @@ impl SnapshotFile {
...
-        be.stream_all::<Self>(p)?
+        Ok(be
+            .stream_all::<Self>(p)?
             .into_iter()
-            .map_ok(Self::set_id)
-            .filter_ok(filter)
-            .try_collect()
+            .map(|item| item.inspect_err(|err| warn!("Error reading snapshot: {err}")))
+            .filter_map(Result::ok)
+            .map(Self::set_id)
```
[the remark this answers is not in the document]
aawsome: Ah, this is about how the repository format works: for "standard" files like Snapshot, Index, Key, the content is saved using the `Id = SHA256(content)` as filename. So, when you retrieve such a file from the backend, you'll pass the `Id` and the the content which can be then deserialized. So you always have the Pair `(Id, RepoFile)` which you can use to act with. And, of course, the Id cannot be saved as part of the content in any RepoFile as inserting this would modify the content resulting in a different Id. (When you save a RepoFile, you give the content and get the Id as result of the saving process).
 
For `SnapshotFile`, however, the `Id` is typically used as identifier, i.e. the users give (parts of) the Id to retrieve snapshots or want to display the `Id` as information. To make this more convenient, the `id` is an item of `SnapshotFile` struct. But, as with other files, this is not set when saving the `SnapshotFile` nor contained in the content saved in the repository. To still have it, we must set it when loading - and for this, there is `set_id` which converts the tuple `(Id, SnapshotFile)` (with snap.id not set) into a `SnapshotFile` (with snap.id set).
N.B: There is also a similar topic with the `original` Id. ;-)

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -853,16 +869,18 @@ impl SnapshotFile {
         be: &B,
         filter: F,
         p: &impl Progress,
-    ) -> RusticResult<Vec<Self>>
+    ) -> RusticResult<impl Iterator<Item = Self>>
```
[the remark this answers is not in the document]
aawsome: I changed the name to `iter_all_from_backend`.

aawsome:
> The `SnapshotRequest` is still private to `snapshotfile.rs` for now. Tell me if I'm wrong but my understanding is that we plan to use to create APIs for `rustic` commands?

Yes. The question is if we want to just internally use `SnapshotRequest`, i.e. have APIs which require something like `&[&str]` - or if we want to use something like `&[SnapshotRequest]` in the API. I thought about the first, but rethinking it: using the second and a method to parse a `&str` or `&[&str]` into `Vec<SnapshotRequest>` might be an even better API as it allows to treat parse errors differently. But I think we should also do this in a follow-up PR.

PR #417, opened 2025-07-18
backend: Use features for opendal services

aawsome:


aawsome:
@simonsan Can you have a look at this PR? The tests currently fails as we'd need to adapt the CI pipelines to use the `ci` feature instead of all features. I like to have your thoughts first before starting to adapt the pipeline.

PR #420, opened 2025-08-03
feat: Add env argument to the run function of command input

rdelaage:
This PR adds an env variables argument to the run function of command inputs. 

This will be useful for this feature request https://github.com/rustic-rs/rustic/issues/1516

### DIFF
--- a/crates/core/src/archiver/file_archiver.rs
+++ b/crates/core/src/archiver/file_archiver.rs
@@ -132,7 +132,10 @@ impl<'a, BE: DecryptWriteBackend, I: ReadGlobalIndex> FileArchiver<'a, BE, I> {
                             .attach_context("path", path.display().to_string())
                         })?;
 
-                    self.backup_reader(r, node, p)?
+                    self.backup_reader(r, node, p).map_err(|err| {
+                        err.prepend_guidance_line("Error while backing up `{path}`")
+                            .attach_context("path", path.display().to_string())
+                    })?
                 } else {
                     (node, 0)
                 };


crates/core/src/repository/command_input.rs
```diff
@@ -128,13 +129,22 @@ impl CommandInput {
     /// # Errors
     ///
     /// * If return status cannot be read
-    pub fn run(&self, context: &str, what: &str) -> RusticResult<()> {
+    pub fn run(&self, context: &str, what: &str, env: &HashMap<String, String>) -> RusticResult<()> {
```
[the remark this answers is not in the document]
rdelaage: I just used a generic with the same type constraints as the std::process::Command.envs method in a new commit. Thanks for the suggestion.

crates/core/src/repository/command_input.rs
```diff
@@ -128,13 +129,22 @@ impl CommandInput {
...
         debug!("calling command {context}:{what}: {self:?}");
-        let status = Command::new(self.command()).args(self.args()).status();
+
+        let mut cmd = &mut Command::new(self.command());
+
+        cmd = cmd.args(self.args());
+
+        for (key, value) in env.iter() {
+            cmd = cmd.env(key, value);
+        }
```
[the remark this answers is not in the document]
rdelaage: You are right, this does the same job and this is nicer. Thanks!

crates/core/tests/command_input.rs
```diff
@@ -76,7 +77,7 @@ fn run_empty() -> Result<()> {
     let command: CommandInput = "".parse()?;
     dbg!(&command);
     assert!(!command.is_set());
-    command.run("test", "empty")?;
+    command.run("test", "empty", &HashMap::<String, String>::new())?;
```
aawsome, line 80: As `Option` also implement `IntoIterator`, I would find using `None` for an empty iterator pretty neat here...

crates/core/tests/command_input.rs
```diff
@@ -91,7 +92,7 @@ fn run_deletey() -> Result<()> {
 
     let command: CommandInput = format!("rm {}", filename.to_str().unwrap()).parse()?;
     assert!(command.is_set());
-    command.run("test", "test-call")?;
+    command.run("test", "test-call", &HashMap::<String, String>::new())?;
```
aawsome, line 95: Maybe use also one example which really sets an env variable?

aawsome (commented):
just two nitpicks about the tests from my side..

crates/core/tests/command_input.rs
rdelaage: This has been done in the last commit

crates/core/tests/command_input.rs
```diff
@@ -1,5 +1,6 @@
 #![allow(missing_docs)]
 #[cfg(not(windows))]
```
aawsome, line 2: I think this should restrict line 4, not line 3

PR #422, opened 2025-08-09
feat: Add fixed size chunking and allow fine-tune of rabin chunking

aawsome:
Adds following extensions to the in-repo config file and implements it accordingly:
- rabin chunker allows to set minimum, maximum and average chunk size
- additionally possible to use fixed-size chunking with a configurable size

This PR implements https://github.com/restic/restic/pull/5470 which specifies these options in the restic design document.


aawsome:
In case that the restic specification won't be accepted or will be delayed, I'd vote for merging this PR nevertheless. 

If then in future restic comes up with a different specification, we can just do the following:
- allow reading both specifiations and process accordingly
- change the default written specifiation to the restic one to keep compatibility
- Users then can always run a `rustic config --set-chunker --set-chunk-size ...` on an existing repo thus turning it into a repo which can be also read by restic.

PR #424, opened 2025-08-21
fix: Add example for time format

aawsome:
closes https://github.com/rustic-rs/rustic/issues/1528


aawsome:
@kapitainsky Thanks for the hint! is changed.

PR #438, opened 2025-11-09
feat!: Add support for batch warmup

philipmw:
Add support for batch warmup

This implements the idea in https://github.com/rustic-rs/rustic/discussions/1430

To use this feature, I wrote a proof-of-concept *warmup-s3-archives* program:
https://gitlab.com/philipmw/warmup-s3-archives

Changes:

* new terminology: "singular" vs "plural" IDs and packs
* add `--warm-up-batch <N>` parameter
* add variables to `--warm-up-command` parameter to support singular and plural IDs and paths
* add `warmup_path()` to the backend interface that provides the S3 key (or other
    key usable by the warmup command) instead of pack ID.

Tested:

* unit tests;
* invoking the warmup program with "%id", "%ids", "%pack", "%packs", and batch
    size of 2 for a total restore size of 3 packs, verifying that the warmup
    command is invoked either separately per ID/pack or with two IDs/packs for
    the first invocation and with one ID/pack for the second invocation.
* killing and restarting restore; this works as long as the warmup program
    is idempotent (which works for S3)
* having the warmup command exit with an error code; rustic aborts the restore
    and prints a correct error message
* running multiple restore operations for different packs in parallel. The
    warmup program ignores notifications for packs that it does not recognize,
    leaving them in the queue and letting another warmup instance process them.

Known limitations and my thoughts for improvement opportunities:

* setup is non-trivial, between the AWS infrastructure and the warmup program configuration.
    It requires some AWS experience and cold storage motivation. But IMO all the
    complexity is specific to the domain; none is incidental.
* rustic does not pass the backend credentials to the warmup program. The warmup
    program is responsible for finding credentials on its own.
    Probably the best solution, at least for AWS, is for both rustic and the
    warmup program to use a common AWS credential provider.
* rustic's progress bar does not reflect warmup progress within a batch; only
    progress of entire batches. There is no protocol for communicating progress
    from a single invocation of the warmup command.
* rustic's warmup parameters are growing in complexity and could use a refactor
    as we discover and clarify cold storage backup scenarios.
    The distinction between `--warm-up-command` and `--warm-up-wait-command`
    seems too subtle. `--warm-up-wait` is too inflexible (since cold storage
    backends' estimates are measured in hours) and can be avoided entirely.

### DIFF
--- a/crates/core/tests/warm_up.rs
+++ b/crates/core/tests/warm_up.rs
@@ -1,0 +1,723 @@
+//! Tests for warm-up batch functionality
+
+use std::{
+    fs::{self, File},
+    io::Read,
+    path::Path,
+    sync::Arc,
+};
+
+use anyhow::Result;
+use rstest::rstest;
+use serde::{Deserialize, Serialize};
+use tempfile::tempdir;
+
+use rustic_core::{
+    CommandInput, Id, NoProgressBars, RepositoryBackends, RepositoryOptions, repofile::PackId,
+};
+use rustic_testing::backend::in_memory_backend::InMemoryBackend;
+
+type RepoOpen = rustic_core::Repository<NoProgressBars, rustic_core::OpenStatus>;
+
+// Test constants
+const DEFAULT_BATCH_SIZE: usize = 1;
+const PACK_ID_HEX_LENGTH: usize = 64;
+
+/// Helper to create a test script that logs invocations and arguments
+/// Returns the tempdir (to keep it alive) and the command
+///
+/// # Arguments
+/// * `log_dir` - Directory where log files will be recorded (each process writes to its own file)
+/// * `exit_code` - Exit code the script should return (0 for success, non-zero for failure)
+#[cfg(not(windows))]
+fn create_test_script_with_exit_code(
+    log_dir: &Path,
+    exit_code: i32,
+) -> Result<(tempfile::TempDir, CommandInput)> {
+    let dir = tempdir()?;
+    let script_name = if exit_code == 0 {
+        "test_warm_up.sh"
+    } else {
+        "test_warm_up_fail.sh"
+    };
+    let script_path = dir.path().join(script_name);
+    let log_dir_path = log_dir.to_string_lossy();
+
+    let exit_line = if exit_code == 0 {
+        String::new()
+    } else {
+        format!("# Exit with error\nexit {exit_code}\n")
+    };
+
+    let script_content = format!(
+        r#"#!/usr/bin/env bash
+# Log that the script was called (each process writes to its own file to avoid interleaving)
+LOG_FILE="{log_dir_path}/warmup_$$.log"
+echo "CALL" >> "$LOG_FILE"
+# Log the number of arguments
+echo "ARGC:$#" >> "$LOG_FILE"
+# Log each argument
+for arg in "$@"; do
+  echo "ARG:$arg" >> "$LOG_FILE"
+done
+{exit_line}
+"#,
+    );
+
+    // Write the script and sync to disk to avoid "Text file busy" errors
+    {
+        use std::io::Write;
+        let mut file = File::create(&script_path)?;
+        file.write_all(script_content.as_bytes())?;
+        file.sync_all()?;
+        // Explicitly drop the file handle before setting permissions
+        drop(file);
+    }
+
+    // Make script executable
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let mut perms = fs::metadata(&script_path)?.permissions();
+        perms.set_mode(0o755);
+        fs::set_permissions(&script_path, perms)?;
+
+        // Sync the parent directory to ensure metadata changes are committed
+        if let Some(parent) = script_path.parent() {
+            if let Ok(dir_file) = File::open(parent) {
+                let _ = dir_file.sync_all();
+            }
+        }
+    }
+
+    let command: CommandInput = script_path.to_string_lossy().to_string().parse()?;
+
+    Ok((dir, command))
+}
+
+/// Helper to create a test script that succeeds
+#[cfg(not(windows))]
+fn create_test_script(log_dir: &Path) -> Result<(tempfile::TempDir, CommandInput)> {
+    create_test_script_with_exit_code(log_dir, 0)
+}
+
+/// Helper to create a test script that fails (exits with non-zero status)
+#[cfg(not(windows))]
+fn create_failing_script(log_dir: &Path) -> Result<(tempfile::TempDir, CommandInput)> {
+    create_test_script_with_exit_code(log_dir, 1)
+}
+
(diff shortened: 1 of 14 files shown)


aawsome:
HI @philipmw!
Are you still interested in working on this? If not, I would try to adapt this in order to get it into rustic!
When thinking about it, I wondered if it was easier to work if we not use `argv` vs `anchor` mode and `packid` vs `path` type, but  just different variables to substitute for all cases. Something like `id` (anchor+packid), `ids` (argv+packid), `path` (anchor+path) and `paths` (argv+packid) (+ additionally `type` solving the mentioned problem about not only warming-up packfiles). What do you think about it?
Sorry again for the long delay after your PR!

philipmw:
Hi, @aawsome , now it's my turn to apologize for taking so long to reply to your feedback. Now that the holidays are over, I will try to reply and act faster on this, as I am still motivated to get it built. I was sitting with my computer already analyzing your feedback when you wrote the most recent update. Here are my responses to each point:

> First I must say that the design to only warm-up pack files was wrong (from my side) - as there are use cases (when repairing a hot/cold repo) where we may want to also warm-up other files. I think we should extend the warmup. This can go in a future PR; however, I think we must keep an eye to ensure this can be easily extended. For backend_path this should be the case, but the name pack_id should be id and then we'll need a way to transfer the type to the command....

Suggested task is to rename `pack_id` to `id`. I have no concerns with this and happy to implement.

> As a batch is now added for "anchor" mode as well, I think this should not be implemented by waiting for each command to finish, but instead spawn all commands of the batch and then wait for all of them - making this also a parallel warm-up.

Change` sequential warmup to parallel. I think this is a good idea, although I am not certain that it would be backward compatible for all current users. Do you have any concerns here?

> For "argv" mode, I find it irritating that the args given for the command are completely ignored. What about keeping them and just append the args of what to warm-up there? (actually that could be also the solution to how to give the type of the file: In the existing args there could be a %type or something like this to be replaced like in "id" mode; just some thoughts I'm having while writing this...)

Can you clarify this feedback? What "args given for the command" are you referring to?

> I must say I was personally a bit confused about the name "anchor" and would maybe call it "variable" or something like this. Do you have another suggestion? Also "argv" is quite technical, maybe "args" is a better name?

I agree, "anchor" is not intuitive. With this name, I was alluding to the [HTML anchor element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a).

Another parallel is Firefox's [dynamic bookmarks](https://support.mozilla.org/en-US/kb/bookmarks-firefox#w_how-to-use-keywords-with-bookmarks) feature: "Wherever the string %s appears in the bookmark's URL, it will be replaced with any words typed in the address bar after the bookmark's keyword and a space, properly URL-encoded, so they can be used as query string parameters to a search engine, for example."

Perhaps another good name for this could be` "substitute", "replace", "dynamic". "Variable" is also fine. Given this list, which name do you like most? (Naming is hard.)

I agree that "argv" could be better renamed to "args". Happy to rename it.

> When thinking about it, I wondered if it was easier to work if we not use argv vs anchor mode and packid vs path type, but just different variables to substitute for all cases. Something like id (anchor+packid), ids (argv+packid), path (anchor+path) and paths (argv+packid) (+ additionally type solving the mentioned problem about not only warming-up packfiles). What do you think about it?

You are suggesting that, instead of `--warm-up-pack-id-input` and `--warm-up-input-type` parameters, each with two possible values, we create a more generic single parameter with four possible values, such as `--warmup-mode`. I don't have a strong opinion here; both ways make sense to me. Happy to implement it either way.

> Also note that there are rustfmt and clippy checks failing - we are quite strict about formatting and clippy-compliance, so these findings must also be fixed (but after discussing the general points..)

Thanks. I didn't realize because the CI step didn't run automatically. Now that it ran for this PR, I will fix the issues.

aawsome:
> Suggested task is to rename `pack_id` to `id`. I have no concerns with this and happy to implement.

I agree that `id` is better suited.

> Change` sequential warmup to parallel. I think this is a good idea, although I am not certain that it would be backward compatible for all current users. Do you have any concerns here?

If `warm-up-batch` is not set, the behavior won't change. So it's just a new feature.
  
> Can you clarify this feedback? What "args given for the command" are you referring to?

Sorry, for me it looked like you were omitting the given args, but https://github.com/rustic-rs/rustic_core/pull/438/files#diff-de508849190b7987f41c9d008e5e4bd90aad3c464a04bf5a73f661cf952b4a62R188 does include them. My fault.


> You are suggesting that, instead of `--warm-up-pack-id-input` and `--warm-up-input-type` parameters, each with two possible values, we create a more generic single parameter with four possible values, such as `--warmup-mode`. I don't have a strong opinion here; both ways make sense to me. Happy to implement it either way.

Actually, I suggest to not use `--warm-up-pack-id-input` and `--warm-up-input-type` at all, but only `--warm-up-batch` and decide on the variables given in the command. Some examples:
- `warm-up-command = "echo %id"` would span [BATCH_SIZE] `echo` commands with each a single id
- `warm-up-command = "echo %ids"` would run a single `echo` command with [BATCH_SIZE] ids given as args
- `warm-up-command = "echo %path"` would span [BATCH_SIZE] `echo` commands with each a single path
- `warm-up-command = "echo %paths"` would run a single `echo` command with [BATCH_SIZE] ids given as args

There is some validation needed so that
- `warm-up-command = "echo %id %path"` would span [BATCH_SIZE] `echo` commands with each a single id and path
- `warm-up-command = "echo %ids %paths"`  would run a single `echo` command with [BATCH_SIZE] ids and [BATCH_SIZE] paths given as args (no idea if this makes sense...), but
- `warm-up-command = "echo %ids %path"` would error out - not clear what to do here!

aawsome (changes requested):
Looks great already! I found some things I think we can simplify the code at some places without changing functionality, so don't be shocked by the number of comments ;-)

philipmw:
@aawsome , I implemented all the changes.

I also noticed that the CI tests are failing for macOS. For one of them, I made a fix, but the other one, I am not certain. It may be already fixed with the latest changes, so let's see how the latest CI build runs.

It seems that the CI won't run automatically til you approve it; is there any way to make it run automatically so that I get faster feedback from it?

crates/core/tests/warm_up.rs
```diff
@@ -0,0 +1,717 @@
...
+    batch_size: usize,
+) -> Result<rustic_core::Repository<NoProgressBars, ()>> {
+    let be = InMemoryBackend::new();
+    let be = RepositoryBackends::new(Arc::new(be), None);
+    let options = RepositoryOptions::default()
+        .password(TEST_PASSWORD)
+        .warm_up_command(command)
+        .warm_up_batch(batch_size);
+
+    rustic_core::Repository::new(&options, &be).map_err(Into::into)
```
aawsome: This is now failing due to the refactoring in #468.

aawsome (approved):
LGTM

PR #448, opened 2025-12-21
feat: Optimize prune rewrite

aawsome:
`prune` used to read each single still needed blob for packs-to-rewrite. While this is OK for cached pack files (tree pack files), it may lead to many API calls when rewriting data packs on remote storages.

This PR optimizes the rewrite: It collects blobs which can be read together and then saves them all from one backend read.

Moreover, the pruning algorithm has been a bit overworked; it now needs less mutex locks. and the steps between rewriting index and repacking are more clearly separated.

### DIFF
--- a/crates/core/tests/integration/prune.rs
+++ b/crates/core/tests/integration/prune.rs
@@ -1,9 +1,11 @@
 use anyhow::Result;
+use bytesize::ByteSize;
 use jiff::Span;
 use rstest::rstest;
 
 use rustic_core::{
-    BackupOptions, CheckOptions, LimitOption, PathList, PruneOptions, repofile::SnapshotFile,
+    BackupOptions, CheckOptions, ConfigOptions, LimitOption, PathList, PruneOptions,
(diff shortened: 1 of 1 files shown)


aawsome:
@mro68 Have you been able to take a look at this PR? I'm especially interested in the constants `LIMIT_PACK_READ` and `MAX_HOLESIZE` - do you think these are reasonable constants?

mro68:
@aawsome I'm still in the midst of the 'wonderful' holiday rush. Christmas, New Year's, kids complaining about the wrong gifts...

I'll be able to make a well-founded statement by the beginning of next year at the latest. At the moment, I can only say that according to my tests, your approach is about 40% slower than my proposed caching, but of course MUCH more elegantly solved. And this increase in speed definitely puts rustic in the “practical” category for GDrive ;-)

PS: If you speak German (which I assume you do), I'll be happy to send you my email address. At the moment, it's git->deepl->git... Feasible, but time-consuming. 

I wish you and your family a Happy and healthy New Year!

PR #468, opened 2026-01-29
feat!: Add masterkey as credential option

aawsome:
Adds the possibility to use a master key directly as credential to open/initialize a repository.

Using the masterkey has the following advantages:
- no need anymore to save the masterkey (encrypted) in the repository. This increases security as it eliminates possible master key leaks due to insecure passwords or a vulnerability in the `scrypt` algorithm.
- by not using the `scrypt` algorithm, opening a repository is much faster and needs much less resources (CPU/Memory) making this option interesting for some use cases. 

Advantages of the still supported password credential are the possibility to backup the masterkey in the repository - and of course allowing to access the repo using a noticeable password.

Note that this change allows to use repositories with only the masterkey, but additionally allows to access a "normal" repository with keyfiles with either a password or the masterky.

As a side effect, most integration tests are now much faster as most now use the masterkey.

This is a breaking change as it changes `Repository` methods.

### DIFF
--- a/crates/core/src/commands/init.rs
+++ b/crates/core/src/commands/init.rs
@@ -13,7 +13,7 @@ use crate::{
     error::RusticResult,
     id::Id,
     repofile::{ConfigFile, KeyId, configfile::RepositoryId},
-    repository::Repository,
+    repository::{Repository, credentials::Credentials},
 };
 
 /// Initialize a new repository.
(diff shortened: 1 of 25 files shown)


s-leroux:
Great job!

I assume the key should be given (unencrypted) as a new global option, isn't it?

Many tools accept command-line arguments, environment variables, and password files (with proper permission check, eg, 600). I don't know if this is a requirement here (or maybe you already implemented that!-)

> Advantages of the _still supported_ password credential are the possibility to backup the masterkey in the repository

(emphasis mine)

I think the password credentials with the master key encrypted in the repository should remain the default strategy. It is reasonably safe, depending on the SLA of the underlying storage provider.  Can the user change the strategy after the repo init?

PR #481, opened 2026-02-22
fix: race condition in warmup tests

### LINKED ISSUE #480: warm-up tests sometimes fail in CI-pipeline
See e.g. https://github.com/rustic-rs/rustic_core/actions/runs/22284226371/job/64459654915?pr=479

The pipeline fails with:
```
failures:

---- test_warm_up_command_failure::case_1 stdout ----

thread 'test_warm_up_command_failure::case_1' (15316) panicked at crates/core/tests/warm_up.rs:637:5:
Error should indicate command failure: `rustic_core` experienced an error related to `running an external command`.

Message:
Error in spawning warm-up command `/tmp/.tmpgufMQc/test_warm_up_fail.sh '%id'`.


Some additional details ...

Context:
- id: 0000000000000000000000000000000000000000000000000000000000000000
- type: WarmUp

Caused by:
Text file busy (os error 26)


Backtrace:
disabled backtrace (set 'RUST_BACKTRACE="1"' environment variable to enable)

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_warm_up_command_failure::case_1

test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.10s

error: test failed, to rerun pass `-p rustic_core --test warm_up`

```

Note this doesn't always happen, seem to be some kind of race condition for parallel test runs..
I never experienced this when running the tests locally on my laptop..

The typical work-around is to just reran the failing pipeline job, however this is a bit annoying..

philipmw:
There is a race condition caused by unit tests creating a script on the filesystem and trying to execute it too fast since creation; the operating system is not allowing execution yet. We don't know how long to wait, but waiting longer will solve it.

closes #480

### DIFF
--- a/crates/core/src/repository/warm_up.rs
+++ b/crates/core/src/repository/warm_up.rs
@@ -1,5 +1,8 @@
+use std::io;
 use std::process::Command;
 use std::thread::sleep;
+
+use backon::{BlockingRetryable, ExponentialBuilder};
 
 use itertools::Itertools;
 use log::{debug, error, warn};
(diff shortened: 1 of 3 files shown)


crates/core/src/repository/warm_up.rs
```diff
@@ -13,10 +15,32 @@ use crate::{
 };
 
 pub(super) mod constants {
-    /// The maximum number of reader threads to use for warm-up.
     pub(super) const MAX_READER_THREADS_NUM: usize = 20;
 }
 
+const MAX_RETRIES: u32 = 5;
```
aawsome: can you please put the constants into `mod constants`

crates/core/src/repository/warm_up.rs
```diff
@@ -13,10 +15,32 @@ use crate::{
...
+const INITIAL_DELAY_MS: u64 = 10;
+
+fn spawn_with_retry<F, T>(mut f: F) -> io::Result<T>
```
aawsome: I like the idea. But can we use `backon` as in https://github.com/rustic-rs/rustic_core/blob/main/crates/backend/src/rest.rs#L113 to perform the actual retrying?

crates/core/src/repository/warm_up.rs
philipmw: done

aawsome (approved):
LGTM! Thanks @philipmw

PR #493, opened 2026-03-28
feat!: scope destination scan to snapshot-relevant paths for restore

srevn:
When restoring to a destination that already contains files (e.g., `rustic restore <snap> /Volumes/Macintosh/`), the destination scan walks the **entire** destination tree before writing anything. This causes two problems:

1. **Permission errors on unrelated directories** — the walk enters directories the user may not have permission to read (e.g., `/System`, `/private`), causing the restore to fail even in `--dry-run` mode.
2. **`--delete` affects unrelated files** — because the snapshot only has sparse entries in ancestor directories (single path components leading to the backed-up content), the merge-join treats everything else in those directories as "additional" and tries to delete it.

This PR replaces the `ignore::WalkBuilder` full-tree walk with a `walkdir::WalkDir` walk that applies two layers of filtering:

1. **Descent filtering** — only enter directories that are part of the snapshot or are ancestors leading to snapshot directories. Irrelevant subtrees are skipped entirely via `skip_current_dir()`, avoiding permission errors and unnecessary I/O.

2. **Content scope filtering** — only yield entries to the merge-join if their parent directory is "within content scope" (the parent or an ancestor directly contains non-directory entries in the snapshot). Entries in pass-through ancestor directories are only yielded if they are navigation targets, preventing `--delete` from removing unrelated files in directories the snapshot only traverses as path components.

### Behavioral changes

- **Without `--delete`:** Restoring to a broad destination (e.g., a whole volume) no longer fails with permission errors on unrelated directories. The "additional entries" warning and stats now only reflect entries within the snapshot's content scope.
- **With `--delete`:** Only files within directories the snapshot fully describes are eligible for deletion. Files in pass-through ancestor directories (e.g., `.VolumeIcon.icns` at the volume root) are no longer touched.

### DIFF
--- a/crates/core/src/commands/restore.rs
+++ b/crates/core/src/commands/restore.rs
@@ -4,16 +4,11 @@ use derive_setters::Setters;
 use jiff::Timestamp;
 use log::{debug, error, info, trace, warn};
 
-use std::{
-    cmp::Ordering,
-    collections::BTreeMap,
-    path::{Path, PathBuf},
-    sync::Mutex,
-};
+use std::{cmp::Ordering, collections::BTreeMap, path::PathBuf, sync::Mutex};
 
-use ignore::{DirEntry, WalkBuilder};
 use itertools::Itertools;
 use rayon::ThreadPoolBuilder;
+use walkdir::WalkDir;
 
 use crate::{
     backend::{
@@ -157,21 +152,20 @@ pub(crate) fn collect_and_prepare<S: IndexedFull>(
     let mut stats = RestoreStats::default();
     let mut restore_infos = RestorePlan::default();
     let mut additional_existing = false;
-    let mut removed_dir = None;
 
-    let mut process_existing = |entry: &DirEntry| -> RusticResult<_> {
+    let mut process_existing = |entry: &walkdir::DirEntry| -> RusticResult<_> {
         if entry.depth() == 0 {
             // don't process the root dir which should be existing
             return Ok(());
         }
 
         debug!("additional {}", entry.path().display());
-        if entry.file_type().unwrap().is_dir() {
+        if entry.file_type().is_dir() {
             stats.dirs.additional += 1;
         } else {
             stats.files.additional += 1;
         }
-        match (opts.delete, dry_run, entry.file_type().unwrap().is_dir()) {
+        match (opts.delete, dry_run, entry.file_type().is_dir()) {
             (true, true, true) => {
                 info!(
                     "would have removed the additional dir: {}",
@@ -185,17 +179,8 @@ pub(crate) fn collect_and_prepare<S: IndexedFull>(
                 );
             }
             (true, false, true) => {
-                let path = entry.path();
-                match &removed_dir {
-                    Some(dir) if path.starts_with(dir) => {}
-                    _ => match dest.remove_dir(path) {
-                        Ok(()) => {
-                            removed_dir = Some(path.to_path_buf());
-                        }
-                        Err(err) => {
-                            error!("error removing {}: {err}", path.display());
-                        }
-                    },
+                if let Err(err) = dest.remove_dir(entry.path()) {
+                    error!("error removing {}: {err}", entry.path().display());
                 }
             }
             (true, false, false) => {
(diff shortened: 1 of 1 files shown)


aawsome:
I thought a bit about this.. 

We need to scan the destination in order to determine which files or file parts really need to be restored and in order to remove existing content when `--delete` is given.
For the first, it suffices to only scan "relevant" subdirs, as proposed in this PR.
For the second, it would also suffice to scan "relevant" subdirs and use `remove_dir_all()` for existings dirs not present in the snapshot. This would still keep the present behavior and yes, it might be that `remove_dir_all()` fails, but this is then the wanted behavior.

Besides this (and somewhat OT for this PR), I think we should add the following:

First, I think we should add an option `overwrite` like in `restic restore --overwrite` which allows to overwrite files
- always
- never
- if modified (by checking mtime)
- if changed (by checking content)

(This would deprecate the `--trust-mtime` option)

Second, I think we should allow to add exclude options for the scanning allowing e.g. to use `--git-ignore` which then would not remove git-ignored dirs even when using the `--delete` option. 

About the implementation @srevn I have some points which I think should be improved, but I'll need to take a bit more time to make a more detailed review, so I beg for a bit more patience...

srevn:
The way I see it, the current workflow is optimized for having a few paths as source for backup and restoring those, selectively, on-demand. But if you have a repo which includes many paths from different areas, having to restore each of them to their location to match the content inside the snapshot becomes tedious.

Snapshot paths are absolute, so there is no other way of restoring a rustic profile to original locations unless using root as destination.

aawsome:
About the changes proposed here:

I think it is too complex. I don't like that first all snapshot entries are collected and I did not see why using `relevant_dirs` and `content_roots` brings much benefits. In future we may refactor metadata processing such that metadata is set for each finished file/dir. Then this `content_roots` may prevent setting correct metadata for empty dirs..

I propose to do it like this:
- In the comparison loop where snapshot entries are compared with the destination entries, call `skip_current_dir()` if the entry is a dir which does not exist. (This may need to change the iterator as you did)
- in process_existing use `remove_dir_all()` instead of `remove_dir()`.

This should result in a non-breaking change which only affects few places but still achieves the main goal that paths not present in the snapshot are not scanned anymore.

srevn:
Reworked based on your feedback. Now the approach is much simpler:

- Replaced `ignore::WalkBuilder` with `walkdir::WalkDir` to get access to `skip_current_dir()`
- In the merge-join loop, when a destination dir is not present in the snapshot, `skip_current_dir()` prevents descending into it — this avoids permission errors and unnecessary I/O
- `remove_dir_all()` was already used by `dest.remove_dir()`, so `--delete` behavior is unchanged
- Removed the `removed_dir` tracking since we no longer descend into extra dirs
- No pre-collection, no HashSets, no behavioral changes — just the scan narrowing

crates/core/src/commands/restore.rs
```diff
@@ -288,20 +251,21 @@ pub(crate) fn collect_and_prepare<S: IndexedFull>(
...
-                error!("Error during collection of files: {err:?}");
+        .sort_by_file_name()
+        .into_iter();
+
+    let next_entry = |walker: &mut walkdir::IntoIter| -> Option<walkdir::DirEntry> {
+        loop {
+            match walker.next()? {
+                Ok(entry) => return Some(entry),
+                Err(err) => error!("Error during collection of files: {err:?}"),
             }
```
aawsome: Is there a problem using `inspect` and `filter_map` like it was in the old version? If not, I'd prefer to keep this. (removes this loop)

crates/core/src/commands/restore.rs
```diff
@@ -311,25 +275,37 @@ pub(crate) fn collect_and_prepare<S: IndexedFull>(
 
             (Some(destination), None) => {
                 process_existing(destination)?;
-                next_dst = dst_iter.next();
+                // don't descend into extra dirs
+                if destination.file_type().is_dir() {
+                    walker.skip_current_dir();
+                }
+                next_dst = next_entry(&mut walker);
```
aawsome: I wonder if it makes sense to modify `process_existing()` to simply return `next_dst`. We have this `skip_current_dir()` in 3 places currently which is kind of ugly (and error-prone)

crates/core/src/commands/restore.rs
```diff
@@ -170,21 +152,20 @@ pub(crate) fn collect_and_prepare<S: IndexedFull>(
     let mut stats = RestoreStats::default();
     let mut restore_infos = RestorePlan::default();
     let mut additional_existing = false;
-    let mut removed_dir = None;
 
-    let mut process_existing = |entry: &DirEntry| -> RusticResult<_> {
+    let mut process_existing = |entry: &walkdir::DirEntry| -> RusticResult<_> {
```
aawsome: Can you please change this and use `use walkdir::DirEntry`?

aawsome (changes requested):
Thanks a lot @srevn!
Looks now much cleaner! I have only a few nitpicks. Besides this, it should be ready!

srevn:
Addressed all three:

- `next_entry` now uses `.inspect(…).find_map(Result::ok)` — clippy suggested `find_map` over `filter_map().next()`, same intent
- `process_existing` takes the walker and returns the next entry, consolidating skip + advance in one place. Three call sites became one-liners
- `DirEntry` imported via `use walkdir::{DirEntry, WalkDir};`

aawsome (approved):
LGTM

ltratt:
This PR seems to be the fix for a completely separate bug I encounter with 0.11.2 where `restore --delete` doesn't delete files in some situations. I _think_ the cause is that directories that are included in an upper (potentially outside the snaphot dir?) `.gitignore` are not traversed (by the `ignore` crate). However, while I can absolutely see this happening on a real-world backup, I can't reproduce this on a small test case, so my reasoning might be faulty (the directory that causes me troubles also happens to have a `.` prefix so maybe that's the case, but again I cannot reproduce this on a small example)! Still, I'm happy that the files I want to be deleted are now being deleted!

PR #524, opened 2026-06-23
feat: stdio protocol between warmup program and rustic (#1782)

philipmw:
[JSON Lines](https://jsonlines.org). Each line must be a single JSON object. The object must contain at least a `type` field that identifies the message.

```json
{"type":"pack-progress","warm":42}
```

A `pack-progress` message tells *rustic* how many of the packs in the current command invocation are expected to be warm now.

| Field | Type    | Meaning |
|-------|---------|---------|
| `type` | string | Must be exactly `"pack-progress"`. |
| `warm` | integer | Number of packs from this invocation expected to be warm. |

*rustic* advances its shared progress bar using these updates.

`warm` is **monotonically non-decreasing** within a single invocation. If your program reports a lower value than a previous report, the lower value is ignored.

If your command exits successfully but never emits a `pack-progress` message, *rustic* still counts the whole invocation as complete. If it reports some progress but the final `warm` is less than the number of packs in the invocation, *rustic* counts the remaining packs as done when the command exits.

So the protocol is strictly an enhancement: emitting it gives users accurate progress; omitting it preserves the pre-protocol behavior of jumping from 0% to 100%.

*rustic* now captures and parses warmup program's stdout. This is a change from older versions, where the warm-up command's stdout was inherited and printed directly to the terminal.

Any stdout line that is not valid JSON is logged by *rustic* at info level, prefixed with `[warmup]`, so that it is still available for debugging. Empty lines are ignored.

I plan to update rustic's documentation with this, once this change is accepted.

### DIFF
--- a/crates/core/src/repository/warm_up.rs
+++ b/crates/core/src/repository/warm_up.rs
@@ -1,12 +1,13 @@
-use std::io;
-use std::process::Command;
-use std::thread::sleep;
+use std::io::{self, BufRead, BufReader};
+use std::process::{Command, Stdio};
+use std::thread::{self, sleep};
 
 use backon::{BlockingRetryable, ExponentialBuilder};
 
 use itertools::Itertools;
-use log::{debug, error, warn};
+use log::{debug, error, info, warn};
 use rayon::ThreadPoolBuilder;
+use serde::Deserialize;
 
 use crate::{
     CommandInput, Id, Progress,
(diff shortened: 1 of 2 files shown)


philipmw:
@aawsome plz

crates/core/src/repository/warm_up.rs
```diff
@@ -212,31 +277,48 @@ fn warm_up_batch_singular(
...
+    let mut readers = Vec::with_capacity(children.len());
+    for (mut child, id) in children {
+        let stdout = child.stdout.take().expect("stdout was piped");
+        let progress = progress.clone();
+        let handle = thread::spawn(move || {
+            // Singular mode handles one pack per command instance.
+            read_progress_output(stdout, 1, &progress)
+        });
+        readers.push((child, id, handle));
+    }
```
aawsome: Why not create the reader handle directly after the child/id has been created, i.e. in the iterator above?

crates/core/src/repository/warm_up.rs
```diff
@@ -250,11 +332,13 @@ fn warm_up_batch_singular(
             .attach_context("type", format!("{ty:?}"))
         })?;
 
+        let current = handle.join().unwrap();
```
aawsome: Can we handle the error here? In rustic_core there should be actually no panic (even though we already do have some panics currently..)

crates/core/src/repository/warm_up.rs
```diff
@@ -325,22 +409,46 @@ fn warm_up_batch_plural(
...
+        RusticError::with_source(
+            ErrorKind::ExternalCommand,
+            "Error waiting for warm-up command `{command}`.",
+            err,
+        )
+        .attach_context("command", command.to_string())
+        .attach_context("type", format!("{ty:?}"))
+    })?;
+
+    let current = handle.join().unwrap();
```
aawsome: see above

aawsome (commented):
Hi @philipmw!

Sorry for the long delay! I had too much family issues in the past months, so I am a bit struggling to catch up..

The PR is mostly fine from my side, just a few things I think can be improved..
Please tell me if you want me to change these!

crates/core/src/repository/warm_up.rs
philipmw: Done.

crates/core/src/repository/warm_up.rs
philipmw: Done.

crates/core/src/repository/warm_up.rs
philipmw: Done.

philipmw:
> Hi @philipmw!
> 
> Sorry for the long delay! I had too much family issues in the past months, so I am a bit struggling to catch up..
> 
> The PR is mostly fine from my side, just a few things I think can be improved.. Please tell me if you want me to change these!

Welcome back! Thanks for the comments. I revised the PR.

aawsome (approved):
I did some small code cleaning. LGTM now! Thanks a lot @philipmw

philipmw:
Thanks, @aawsome ! Regarding clippy - I know it's not the first time you had to point out lint problems; sorry about that. Can you change the settings to run CI automatically for pull requests? Right now it never runs for me, saying it needs you to start it manually.

PR #531, opened 2026-07-07
feat(packer): upload pack files concurrently

ibash:
The pack file writer (`Actor`) uploaded packs strictly one at a time: it was constructed with `par = 1` and its `par` argument was ignored, so the upload stage was a sequential `.map(|load| fwh.process(load))`. On a high-latency backend link a single connection cannot fill the pipe, so backup/copy/repack throughput was capped regardless of configuration.

Make the writer upload up to N packs concurrently by turning the upload stage into a bounded `parallel_map_scoped_custom` with an explicit thread count (tracks the network, not CPU cores). Only `write_bytes` is parallelized; hashing stays sequential and indexing stays serialized in the final `try_for_each`, so the indexer and `PackerStats` need no new synchronization. pariter yields in input order, so the finalize barrier and first-error propagation are preserved: a snapshot can never reference a pack whose upload has not completed.

Expose the concurrency as `RepositoryOptions::upload_concurrency` (`upload-concurrency` in config, `--upload-concurrency` flag), defaulting to 8, threaded through the archiver, BlobCopier, and tree modifier/rewriter.

Tests (crates/core/tests/integration/upload.rs):
- round-trip (backup -> check --read-data -> restore -> byte compare) across many packs at concurrency 1/4/16
- injected mid-stream upload failure aborts the backup without committing a snapshot

Refs rustic issues #980, #981, #1215.

### DIFF
--- a/crates/core/src/archiver.rs
+++ b/crates/core/src/archiver.rs
@@ -83,13 +83,22 @@ impl<'a, BE: DecryptFullBackend, I: ReadGlobalIndex> Archiver<'a, BE, I> {
         config: &ConfigFile,
         parent: Parent,
         mut snap: SnapshotFile,
+        upload_concurrency: usize,
     ) -> RusticResult<Self> {
         let indexer = Indexer::new(be.clone()).into_shared();
         let mut summary = snap.summary.take().unwrap_or_default();
         summary.backup_start = Zoned::now();
 
-        let file_archiver = FileArchiver::new(be.clone(), index, indexer.clone(), config)?;
-        let tree_archiver = TreeArchiver::new(be.clone(), index, indexer.clone(), config, summary)?;
+        let file_archiver =
+            FileArchiver::new(be.clone(), index, indexer.clone(), config, upload_concurrency)?;
+        let tree_archiver = TreeArchiver::new(
+            be.clone(),
+            index,
+            indexer.clone(),
+            config,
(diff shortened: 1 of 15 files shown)


ibash:
Note: this is written by claude and I haven't reviewed the code yet! Please forgive me for ai slop.

ibash:
Ended up switching to GCS as my backend storage and things are fast again.

aawsome:
Hi @ibash 
Tecnically claude did the right thing here. The concurrency (and buffer lengths of the queue) was already designed to be set from outside, it was just not offered to the users. 
However, there are some problems:
- rustic keeps the packfiles in-memory. So, memory can be a problem here if we need to keep a lot of pack files in-memory. (And the restic solution to use temporary files is not much better as typically those are and should be on a memory-backed tmpfs)
- rustic can use quite large pack files. For those it is better to do a parallel upload of a single file (which is supported by opendal if the backend supports it). Else you get memory problems.
- For the same reason it doesn't make sense to use large queue (size 1 or maybe even 0 should do it). we shouldn't have pack files just lying around. They should be either in-creation or being-processed.

That said, I am currently working on a fully actor-based implementation of the actual processing/reading/writing logic which seems to be quite promising. I'd add some options about reading/writing/compute concurrency there as this will be in a single place, then.

PR #535, opened 2026-07-26
feat: implement some exclude options for opendal source

hacker:
the new opendal as source backend supports no exclusion whatsoever, which is quite a shortcoming (and should probably be documented). This commit partially fixes that, it does not implement all exclusions, but at least it implements glob exclusions, which is better than nothing.

I do not consider this commit a final resolution, especially given the general intention to get rid of dependency on `ignore`, but I'd say it's definitely an improvement.


aawsome (dismissed):
Hi @hacker Thanks for the PR! Indeed once we refactor excludes, this kind of glob matching should go to a central part instead of being implemented per-source. But it is definitively an improvement!

LGTM!

aawsome (approved):
I fixed formatting and clippy issues.
LGTM!

PR #539, opened 2026-08-04
chore(deps): bump opendal to 0.58.1

Xuanwo:
## Summary

Bump the optional `opendal` dependency in `rustic_backend` from `0.57.0` to `0.58.1`.

OpenDAL 0.58.1 requires Rust 1.91, so the workspace MSRV is raised from `1.88.0` to `1.91.0` (including README policy text). Without that, the MSRV CI job (`cargo hack check --rust-version`) would not match the dependency graph.

No application code changes were required: the backend already uses the current OpenDAL APIs (`Operator::via_iter`, blocking `Operator`, `Timestamp::into_inner`, write returning `Metadata`). Existing `services-*` / `layers-throttle` / `blocking` features continue to resolve on 0.58.1. `object_store_opendal` is not used in this repo.

## Validation

- `cargo check -p rustic_backend --features opendal`
- `cargo test -p rustic_backend --features opendal --lib` (32 passed)

## Notes

- Lockfile resolves `opendal` / `opendal-core` and related service/layer crates to `0.58.1`.
- Full monorepo CI (cross/powerset) was not run locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


aawsome (approved):
LGTM! Thanks a lot @Xuanwo

PR #540, opened 2026-08-06
feat(backend/opendal): add opt-in per-part write progress via ProgressLayer + chunked writer PR Description

### LINKED ISSUE #538: Add opt-in real-time upload progress to the OpenDAL backend (per multipart part instead of per completed pack)
> \[!NOTE\] This issue is **not** about backup read/dedup progress.
>
> The read-side progress is already accurate.
>
> The request is to expose **upload progress** during OpenDAL writes so
> that applications can observe upload progress while a pack is still
> being uploaded, instead of only after the entire pack has completed.

------------------------------------------------------------------------

# Background

In an application integrating `rustic_core` + `rustic_backend`, two
independent progress metrics are reported:

  -----------------------------------------------------------------------
  Metric                            Meaning
  --------------------------------- -------------------------------------
  `readBytes` / `readTotal` /       Original source data read and
  `readProgress`                    deduplicated

  `writtenBytes`                    Actual bytes written to the backend
                                    (compressed/encrypted pack bytes)
  -----------------------------------------------------------------------

The read-side metric behaves correctly.

The write-side metric only advances after an entire pack upload has
finished, making upload progress and throughput appear coarse.

------------------------------------------------------------------------

# Observed behavior

Example logs:

``` text
08-02 18:11:49.765 PROGRESS-DIAG readBytes=723421    readTotal=132147200 readProgress=0.0054743574 writtenBytes=0        speed=0
08-02 18:13:25.918 PROGRESS-DIAG readBytes=132147200 readTotal=132147200 readProgress=1.0          writtenBytes=45127834 speed=469336
08-02 18:14:26.931 PROGRESS-DIAG readBytes=132147200 readTotal=132147200 readProgress=1.0          writtenBytes=71746505 speed=436288
08-02 18:14:27.276 PROGRESS-DIAG readBytes=132147200 readTotal=132147200 readProgress=1.0          writtenBytes=71760889 speed=451843
08-02 18:16:23.199 PROGRESS-DIAG readBytes=80384000  readTotal=80384000  readProgress=1.0          writtenBytes=44556589 speed=388682
08-02 18:16:57.092 PROGRESS-DIAG readBytes=80384000  readTotal=80384000  readProgress=1.0          writtenBytes=58243235 speed=403820
```

`readProgress` reaches 100% long before uploads finish, while
`writtenBytes` advances in large steps because it is only updated after
a completed pack upload.

------------------------------------------------------------------------

# Analysis

## 1. Read progress is already correct

The backup size is determined before backup begins
(`p.set_length(size)`), and every processed chunk advances progress via
`p.inc(size)` regardless of whether it is deduplicated or uploaded.

Therefore `readProgress == 100%` already correctly means all source data
has been scanned and deduplicated.

No change is required on the read side.

## 2. Upload progress is limited by the current write path

Current upload path:

``` text
packer
  ↓
save()
  ↓
OpenDALBackend::write_bytes()
  ↓
Operator::write(path, buf)
```

The current implementation uses a whole-object write:

``` rust
self.operator.write(&filename, buf)?;
```

From `rustic_backend`'s perspective there are only two observable
events:

``` text
write begins
...
write returns
```

No intermediate progress information is exposed.

## 3. A Layer alone is insufficient

A custom OpenDAL `Layer` cannot provide incremental progress by itself.

When the complete object is passed to `Operator::write()`, the
underlying writer typically receives only one write operation.
Consequently, the Layer also observes only one successful write.

Incremental progress therefore requires **both**:

1.  A progress-aware Layer.
2.  A chunked write path that feeds the writer incrementally.

These two components are complementary.

## 4. Current extension points

`OpenDALBackend` currently assembles its `Operator` and Layer chain
internally, leaving no public extension point for upload progress
reporting.

------------------------------------------------------------------------

# Proposed enhancement

Provide an **opt-in** upload progress mechanism for the OpenDAL backend.

The implementation should allow:

-   attaching upload progress reporting;
-   using a chunked multipart-capable write path;
-   preserving existing behavior for all current callers unless they
    explicitly opt in.

The exact API is open for discussion.

------------------------------------------------------------------------

# Progress semantics

The reported progress represents bytes successfully accepted by the
underlying OpenDAL writer.

For multipart-capable services, this corresponds approximately to
multipart part submission rather than actual socket transmission.

This is sufficient for progress bars, throughput estimation and
diagnostics.

------------------------------------------------------------------------

# Expected result

Instead of:

``` text
0
45 MiB
71 MiB
118 MiB
```

progress would advance approximately once per multipart part:

``` text
0
8 MiB
16 MiB
24 MiB
...
```

The exact granularity depends on backend implementation and multipart
configuration.

Small objects below the multipart threshold may still complete in a
single step.

------------------------------------------------------------------------

# **Prototype validation**

A prototype implementation combining:

-   a progress-aware OpenDAL Layer;
-   an opt-in chunked write path;

was validated against **Tencent COS**.

Observed output:

``` text
PROGRESS-COS t=1786002936260 counter=0
PROGRESS-COS t=1786002936461 counter=8388608
PROGRESS-COS t=1786002937265 counter=16777216
PROGRESS-COS t=1786002937466 counter=25165824
PROGRESS-COS t=1786002937667 counter=33554432
PROGRESS-COS t=1786002938069 counter=41943040
PROGRESS-COS t=1786002938269 counter=50331648
PROGRESS-COS t=1786002938470 counter=58720256
PROGRESS-COS t=1786002938671 counter=67108864
PROGRESS-COS t=1786002941082 counter=134217728
```

The counter advanced incrementally during upload.

For this prototype:

-   **Object size:** 128 MiB
-   **Chunk size:** 8 MiB
-   **Backend:** Tencent COS
-   **Observed granularity:** one update per multipart part

This demonstrates the feasibility of the proposed approach.

------------------------------------------------------------------------

# Scope

This proposal does **not** change:

-   backup semantics;
-   deduplication;
-   repository format;
-   repository compatibility.

Existing callers keep the current behavior.

Applications that explicitly opt into upload progress reporting may use
a different multipart-capable write strategy to expose incremental
progress.

------------------------------------------------------------------------

# Motivation

The goal is **observability rather than new backup functionality**.

Existing users are unaffected.

Applications that opt in gain accurate upload progress, smoother
throughput reporting and improved diagnostics while maintaining
repository compatibility.

543069760:
## Summary

This PR adds opt-in, real-time write progress reporting to the OpenDAL backend.

Previously, `OpenDALBackend::write_bytes` uploaded each file with a single `operator.write(path, buf)` call. For services backed by an S3-style multipart API (e.g. Tencent COS), OpenDAL's `MultipartWriter` applies a `write_once` optimization when the whole buffer is handed over in one shot: with a known size and a single write, it issues a single `PutObject` request instead of initiating a multipart upload. As a result, any progress observer only ever sees a single jump from 0 to the full object size at the very end of the upload — there is no intra-object progress.

This makes it impossible to surface smooth upload progress (e.g. to a JNI caller driving a progress bar) for large packs.

## Approach

We introduce a custom OpenDAL Layer (`ProgressLayer`) that wraps the underlying writer and atomically accumulates the number of bytes written each time a `Buffer` is flushed down to the inner writer. To actually produce multiple flushes (and therefore multiple progress steps), the backend switches from a single one-shot `operator.write` to a chunked writer that feeds the data in fixed-size chunks, which forces OpenDAL past its `write_once` optimization and into the real multipart path (`InitiateMultipartUpload` + multiple `UploadPart` + `CompleteMultipartUpload`).

The behavior is opt-in and fully backward compatible:
- The existing `OpenDALBackend::new(...)` constructor is unchanged and stores `counter: None`. Backends created this way keep the original single-shot `operator.write(buf)` path — zero behavior change for all existing backends (fs / sftp / rest / s3 / …).
- A new `OpenDALBackend::new_with_progress(...)` constructor accepts an `Option<WrittenCounter>`. Only when a counter is provided is the `ProgressLayer` assembled and the chunked writer path taken.

### `crates/backend/src/progress_layer.rs`
- `pub type WrittenCounter = Arc<AtomicU64>` — shared, cheaply-cloneable counter handle. The caller keeps one `Arc` clone to poll progress.
- `ProgressLayer` — an `opendal::raw::Layer` holding a `WrittenCounter`.
- `ProgressAccessor` / `ProgressWriter` — `LayeredAccess` / `oio::Write` wrappers. `ProgressWriter::write` records `bs.len()` only after the inner write succeeds (so failed writes / retries don't inflate the count), using `Ordering::Relaxed`. `close` / `abort` are passed straight through.
- The layer operates on the async operator (assembled before the blocking wrapper). Granularity = one increment per chunk the underlying service writer receives (i.e. per multipart part for S3-like services).

### `crates/backend/src/lib.rs`
- Export the new module and public types so callers (e.g. the JNI layer) can construct a `WrittenCounter` and call `new_with_progress`.

### `crates/backend/src/opendal.rs`
- Import `opendal::options::WriteOptions` (used to set the writer chunk size).
- `mod constants`: add `pub(super) const CHUNK_SIZE: usize = 8 * 1024 * 1024;` (8 MiB). Chosen to be larger than the ~5 MiB minimum part size required by S3/COS-style services, so it reliably crosses OpenDAL's `write_once` optimization and triggers a real multipart upload.
- `OpenDALBackend` struct: add a `counter: Option<WrittenCounter>` field that records whether a `ProgressLayer` was assembled, so `write_bytes` can branch on it.
- `new(...)`: unchanged logic; now sets `counter: None`.
- `new_with_progress(path, options, counter)`: same operator-construction logic as `new`, but when `counter.is_some()` it assembles the `ProgressLayer` on the async operator (using a cheap `Arc` clone so the counter is also stored in the struct), then wraps it in the blocking `Operator`. Stores the counter in the struct.
- `write_bytes(...)`: branch on `self.counter`:
  - `Some(_)` → build a chunked blocking writer via `operator.writer_options(path, WriteOptions { chunk: Some(constants::CHUNK_SIZE), ..Default::default() })`, loop over `buf.chunks(constants::CHUNK_SIZE)` calling `writer.write(chunk)`, then `writer.close()` (return value discarded to satisfy `-W unused-results`). This drives `ProgressLayer` to fire once per part.
  - `None` → keep the original single-shot `operator.write(&filename, buf)`.

### `crates/backend/tests/`
- `progress_layer.rs` (runs on the memory service, always executed):
  - `progress_layer_counts_written_bytes` — a backend built with a counter reports `counter >= bytes_written` after a write.
  - `default_path_without_counter_still_writes` — regression: with `counter = None` the write works and reads back identically (default path unaffected).
- `progress_layer_cos.rs` (real Tencent COS, `#[ignore]` by default):
  - `cos_progress_layer_per_part_granularity` — requires `COS_SECRET_ID`, `COS_SECRET_KEY`, `COS_BUCKET`, `COS_ENDPOINT`, `COS_ROOT`. Uploads a 128 MiB pack and asserts the counter advances in multiple steps, cleaning up the object afterwards.

## Verification

- `cargo build -p rustic_backend --features opendal` — clean (no new warnings).
- `cargo clippy -p rustic_backend --features opendal` — no new lints introduced (pre-existing repo warnings left untouched).
- `cargo test -p rustic_backend --features opendal` — memory regression tests pass; the COS test is correctly reported as ignored.
- Real COS run (`cargo test -p rustic_backend --features opendal --test progress_layer_cos -- --ignored --nocapture`): the network trace shows `InitiateMultipartUpload` + multiple `UploadPart` (`partNumber=…&uploadId=…`, `content-length: 8388608`) + `CompleteMultipartUpload`, and the `PROGRESS-COS` log shows the counter stepping in 8 MiB increments: `0 → 8388608 → 16777216 → … → 134217728` (16 parts for 128 MiB).

## Backward Compatibility

- Existing `new(...)` callers are unaffected: `counter = None` → original single-shot `operator.write` path, identical behavior for all backends.
- The chunked writer and multipart path are only taken when a caller explicitly opts in via `new_with_progress(...)` with a counter.

## Summary

This PR adds opt-in, real-time write progress reporting to the OpenDAL backend.

Closes #538


aawsome (changes requested):
Hi @543069760, thanks for opening this PR!

This PR combines different topics: Besides the progress topic implemented on opendal, it also adds a cancellation feature (which is also a breaking change).
Can you please focus on the first topic and remove the changes for the cancellation?
This could be a topic for a different PR, but please let's discuss this first in an issue or discussion as at this point I a didn't even get which problem this tries to solve..

Also there are quite some changes in comments which only add empty spaces - please also remove them!

543069760:
@aawsome Hello, I will eliminate other irrelevant changes and then PR again￼￼￼￼￼￼￼￼￼￼￼￼

PR #546, opened 2026-08-25
fix: stop tree workers after receiver closes

magrathean-uk:
## Summary

- stop tree-loading workers when the consumer has already exited
- handle a closed output channel inline without panicking

Related: rustic-rs/rustic#1877, rustic-rs/rustic#1790, rustic-rs/rustic#1437, rustic-rs/rustic#1400

## Validation

- cargo test --locked -p rustic_core --lib: 168 passed
- cargo clippy -p rustic_core --lib --tests --locked -- -D warnings
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/core/src/blob/tree.rs
+++ b/crates/core/src/blob/tree.rs
@@ -636,6 +636,24 @@ pub struct TreeStreamerOnce {
     finished_ids: usize,
 }
 
+fn forward_loaded_trees<F>(
+    in_rx: Receiver<(PathBuf, TreeId, usize)>,
+    out_tx: Sender<RusticResult<(PathBuf, Tree, usize)>>,
+    load_tree: F,
+) where
+    F: Fn(TreeId) -> RusticResult<Tree>,
+{
+    for (path, id, count) in in_rx {
+        if out_tx
+            .send(load_tree(id).map(|tree| (path, tree, count)))
+            .is_err()
+        {
+            // The consumer stopped after an error. No more work can be observed.
+            break;
+        }
+    }
+}
+
 impl TreeStreamerOnce {
     /// Creates a new `TreeStreamerOnce`.
     ///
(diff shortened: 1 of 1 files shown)


crates/core/src/blob/tree.rs
```diff
@@ -636,6 +636,24 @@ pub struct TreeStreamerOnce {
     finished_ids: usize,
 }
 
+fn forward_loaded_trees<F>(
```
aawsome: I don't see why we should extract this into a separate function. Can you add the error handling directly?

crates/core/src/blob/tree.rs
```diff
@@ -949,3 +963,28 @@ pub(crate) fn merge_nodes(
...
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use crossbeam_channel::{bounded, unbounded};
+
+    use super::{Tree, forward_loaded_trees};
+
+    #[test]
+    fn tree_loader_exits_when_output_receiver_is_dropped() {
```
aawsome: For me this unit test is kind of useless. We now test that no panic due to an error occurs. Before we had an `unwrap` which does panic in that case. Instead of the unit test, we could have used a lint to check that no `unwrap` exists (but this doesn't make sense as we currently have still too many `unwrap()`s in the code.

crates/core/src/blob/tree.rs
magrathean-uk: Done in d9f9b0f. The closed-channel handling now lives directly in the worker loop.

crates/core/src/blob/tree.rs
magrathean-uk: Removed the helper-level regression test as suggested. The full core library suite passes: 168/168.

aawsome (approved):
LGTM

PR #547, opened 2026-08-25
fix: run delete hook when replacing a file

magrathean-uk:
## Summary

- preserve the atomic replacement path
- run post-delete after a successful config replacement and before post-create
- check for an existing destination only for config writes with a configured post-delete hook
- cover the create, overwrite, delete, create hook order

Related: rustic-rs/rustic#1417

## Validation

- cargo test --locked -p rustic_backend: 33 passed; doc tests passed
- cargo clippy -p rustic_backend --tests --locked -- -D warnings
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/backend/src/local.rs
+++ b/crates/backend/src/local.rs
@@ -522,7 +522,8 @@ impl WriteBackend for LocalBackend {
 
         trace!("writing tpe: {tpe:?}, id: {id}");
         let filename = self.path(tpe, id);
-        let replaces_existing_file = filename.exists();
+        let replaces_existing_config =
+            tpe == FileType::Config && self.post_delete_command.is_some() && filename.exists();
 
         let parent = self.base_path(tpe, id);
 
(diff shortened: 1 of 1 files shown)


crates/backend/src/local.rs
```diff
@@ -522,6 +522,7 @@ impl WriteBackend for LocalBackend {
 
         trace!("writing tpe: {tpe:?}, id: {id}");
         let filename = self.path(tpe, id);
+        let replaces_existing_file = filename.exists();
```
aawsome: An existing file with different content should actually only happen for the config file (all other files have the SHA256 as filename). I wonder if we should restrict this check to this setting.
Or maybe only check existence in case there is a `post-delete-command` ?

magrathean-uk: Done in c3db209. Existence is now checked only for config writes and only when post-delete-command is configured. The ordering test still covers config replacement.

aawsome (approved):
LGTM

PR #548, opened 2026-08-25
fix: include path in canonicalization errors

### LINKED ISSUE #429: `canonicalizing path failed` does not show which path failed
Hi! I switched several backups from restic to rustic, and I love it, many thanks!

Converting my last repository, which had a lot of backup path, I stumbled across this error:

```
canonicalizing path failed: `Os { code: 2, kind: NotFound, message: "No such file or directory" }`
```

It took me some research to find that this was one of the backup paths, at first I thought this was one of the globs that didn't match.

I think it would make it more explicit and user-friendly to show an error like this instead:

```
canonicalizing path /path/that/not/exists failed: `Os { code: 2, kind: NotFound, message: "No such file or directory" }`
```

What do you think? I can try doing it if you're ok with the proposal.

magrathean-uk:
## Summary

- retain the failing path when canonicalization fails
- include it in the user-facing error and regression test

Fixes #429

## Validation

- cargo test -p rustic_core path_list_sanitize_error_includes_path --locked
- cargo check -p rustic_core --locked
- cargo fmt --check

### DIFF
--- a/crates/core/src/repofile/snapshotfile.rs
+++ b/crates/core/src/repofile/snapshotfile.rs
@@ -46,8 +46,8 @@ pub enum SnapshotFileErrorKind {
     ValueNotAllowed(String),
     /// removing dots from paths failed: `{0:?}`
     RemovingDotsFromPathFailed(std::io::Error),
-    /// canonicalizing path failed: `{0:?}`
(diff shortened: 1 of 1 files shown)


aawsome (approved):
LGTM! Thanks @magrathean-uk

PR #551, opened 2026-08-25
feat: serialize prune statistics

### LINKED ISSUE #223: Add JSON support/impl serialize for prune statistics
[...] I realized that no data structure does `impl Serialize` - so this has to wait until JSON support for the prune statistics has been added.

_Originally posted by @aawsome in https://github.com/rustic-rs/rustic_core/pull/221#discussion_r1590161138_

magrathean-uk:
## Summary

- add Serialize for prune statistic types and BlobTypeMap values
- serialize human-readable debug maps as deterministic key/stats entry arrays, since JSON cannot use structured map keys
- retain native map serialization for non-human-readable serializers

Fixes #223

## Validation

- cargo test -p rustic_core --lib --locked
- cargo clippy -p rustic_core --lib --tests --locked -- -D warnings
- cargo check -p rustic_core --locked

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1254,6 +1254,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6866f3bfdf8207509a033af1a75a7b08abda06bbaaeae6669323fd5a097df2e9"
 dependencies = [
  "enum-map-derive",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


aawsome:
Hi @magrathean-uk,
Thanks for the PR! However I must say that the main point here is not the fact that some structs don't implement `Serialize`. It is more that this currently is an internal structure which is not yet really well-thought in terms of what to provide in an API.
If we add `Serialize`, we consequently should also make the whole stats struct `pub` - but before, we should reconsider what we really want to be pub here and how to structure it...

magrathean-uk:
Agreed. Serializing the current internal shape would make it an API before that API has been designed. I will close this implementation rather than freeze the wrong structure; #223 can be revisited once the intended public output type is defined.

PR #556, opened 2026-08-29
feat!: Add more reqwest options for rest+opendal backends

### LINKED ISSUE #555: Allow OpenDAL backends to configure redirect Referer policy
## Summary

Please allow OpenDAL repositories to opt out of reqwest's automatic `Referer`
header on redirects, without requiring applications to maintain a custom
`rustic_backend` build.

A possible repository option is:

```toml
[repository]
repository = "opendal:webdav"

[repository.options]
http-referer = "false"
```

The name and exact API are open for discussion. The important contract is that
the option is transport policy, not a provider-specific WebDAV setting.

## Current behavior

At rustic_core commit `e5747651419a13bf6d23fbf51a578b6faef8f137`,
`crates/backend/src/opendal.rs` constructs the backend with:

```rust
opendal::Operator::via_iter(scheme, options)
```

This uses OpenDAL's default HTTP transport. reqwest follows redirects by default
and automatically generates `Referer` unless its client is built with
`.referer(false)`.

This matters for a generic WebDAV pattern:

1. The authenticated WebDAV origin receives a ranged GET.
2. It returns a cross-origin 302 to a signed object URL.
3. reqwest correctly removes `Authorization` and cookies and preserves `Range`,
   but adds a cross-origin `Referer`.
4. An object endpoint with a restrictive referer policy rejects the otherwise
   valid signed request.

In a live compatibility probe, the same signed URL returned byte-exact `206`
and full-object data when requested without `Referer`, while stock rustic was
rejected after following the redirect with one. No repository write was needed
to reproduce the difference.

## Suggested direction

- Parse a backend-owned boolean option such as `http-referer` and remove it from
  the service options before dynamic service construction.
- Preserve the current reqwest default when the option is absent.
- When explicitly set, construct a custom reqwest client with
  `ClientBuilder::referer(value)`, wrap it in OpenDAL's `ReqwestTransport`, and
  inject the transport into the operator's `OperationContext`.
- Inject the custom transport before applying `ConcurrentLimitLayer`. OpenDAL
  0.58's layer wraps the HTTP transport and holds its HTTP permit until the
  response body is dropped; replacing the context afterward would bypass that
  body-lifetime guarantee.
- Keep the option generic across HTTP-based OpenDAL services.
- Reject invalid boolean values instead of silently changing policy.

OpenDAL 0.58 already exposes both a custom `ReqwestTransport::new(client)` and
per-operator `OperationContext` HTTP transport injection, so this appears to be
an application configuration/plumbing gap rather than a new redirect
implementation.

## Compatibility and security invariants

- Default behavior remains unchanged.
- Cross-origin `Authorization`, proxy authorization, and cookies remain removed.
- `Range` survives the redirect.
- Disabling automatic `Referer` does not disable redirect limits or TLS checks.
- The option must not be forwarded as an unknown WebDAV/service option.

## Suggested tests

Use two local test servers rather than an external provider:

1. An authenticated WebDAV endpoint returns a cross-origin 302.
2. The target returns 403 when `Referer` is present and `206` when it is absent.
3. Assert default behavior is unchanged.
4. Assert `http-referer=false` succeeds, preserves `Range`, and does not leak
   WebDAV credentials to the target.
5. Assert malformed configuration fails during backend admission.
6. With `connections=1`, assert a second redirected response cannot start until
   the first response body is dropped.

## Non-goals

- No provider-specific Aliyun behavior.
- No change to repository format or backup semantics.
- No change to the default redirect policy.
- No attempt to make provider direct-link 302 part of the S3 protocol.

## Environment and references

- rustic: v0.11.4
- rustic_backend: 0.7.0
- OpenDAL: 0.58.1
- reqwest: 0.13.4
- OpenDAL custom transport documentation:
  https://docs.rs/opendal-http-transport-reqwest/latest/opendal_http_transport_reqwest/
- reqwest `ClientBuilder::referer`:
  https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.referer

## AI assistance disclosure

OpenAI Codex assisted with source tracing, reproduction analysis, and wording.
The proposal and cited current behavior were reviewed against the referenced
source revisions before submission.

Related OpenList proxy reliability proposal:
https://github.com/OpenListTeam/OpenList/issues/2969

### LINKED ISSUE #1895: option --insecure-tls is missing
restic has the option `--insecure-tls` which is missing in rustic:

```
# rustic version
rustic v0.11.4-nightly

# rustic --insecure-tls snapshots
error: unexpected argument '--insecure-tls' found
```

I need this option because of self-signed https certificate.

aawsome:
This PR introduces a reqwest client which is used by both the rest and the opendal backend (for http-based backend access).

Furthermore it adds some options to fine-tune this client:
- `http-insecure-tls` -> allows to use self-signed certificates
- `http-referer` -> allows to configure whether a Referer header is used
- `timeout` (new for opendal) -> allow to configure the timeout for a single http request
- `cacert` (new for opendal)  -> allows to read and use a root certificate from a local file
- `tls-client-cert` (new for opendal) -> allows to read and use a client certificate from a local file.

closes #555
closes https://github.com/rustic-rs/rustic/issues/1895

This is a breaking change as `RestBackend::new` now takes the options as `BTreeMap<String,String>` instead as accepting any iterator.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,6 +264,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "asyncband"
+version = "0.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94a214ba60d6231afd0e805e3c27c45a1626d9debaa5a5061c45a1ea1b2f1ed0"
+dependencies = [
+ "hashbrown 0.17.1",
+ "slab",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2564,9 +2574,9 @@ checksum = "c08d65885ee38876c4f86fa503fb49d7b507c2b62552df7c70b2fce627e06381"
 
 [[package]]
 name = "opendal"
-version = "0.58.1"
+version = "0.58.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f20562cc7447fcc915fc5c23df305a412ea80a733c9f2fd9e2d267e2815be6d"
+checksum = "33dbff14cc9bb085224256d6a81289d2f3202e85b06f408d42534b42162a4231"
 dependencies = [
  "ctor",
  "opendal-core",
@@ -2603,11 +2613,12 @@ dependencies = [
 
 [[package]]
 name = "opendal-core"
-version = "0.58.1"
+version = "0.58.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec75551ff4cf3e57da98979f6a937aaa9ddb3915bf68cc17d03df733be6646ed"
+checksum = "48dbcef97d3eb7591db2c18d5cae95c836bcce07359b98d98dd6f4e861eb77b7"
 dependencies = [
  "anyhow",
+ "asyncband",
  "base64 0.23.1",
  "bytes",
  "futures",
@@ -2615,7 +2626,6 @@ dependencies = [
  "jiff",
  "log",
  "md-5",
- "mea",
  "percent-encoding",
  "quick-xml",
  "reqsign-core",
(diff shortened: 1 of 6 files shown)


aawsome:
I need to think if we also rename the existing options and prefix them with `http-` as this is anyway a breaking change..

aawsome:
Tested this with a local rest and s3 server.

PR #557, opened 2026-08-30
fix: preserve files of all sources in multi-source backup

Lauri-Nomme:
## Problem

Backing up more than one source path in a single call silently drops the second source's files — or panics with `StripPrefixError`.

Repro:
```
rustic_core$ cargo test --test integration test_backup_multi_source_preserves_files_of_all_sources
```
before this change:
```
thread '<unnamed>' panicked at crates/core/src/archiver.rs:163:66:
called `Result::unwrap()` on an `Err` value: StripPrefixError(())
```

Reported upstream: [rustic-rs/rustic#1905](https://github.com/rustic-rs/rustic/issues/1905)

## Root cause

`Archiver::archive` received only `backup_paths[0]` as the single `backup_path` used for `strip_prefix` (`crates/core/src/commands/backup.rs`). The `ignore`-based walker emits **absolute paths per source root** for every entry. With multiple sources, an entry from the second source (`.../src2/file.txt`) cannot `strip_prefix(backup_paths[0])`, so:
- with `--as-path` set: the `unwrap()` panics,
- without `--as-path`: the second source's file blobs never get archived while its directory tree is still written — a silent, successful-looking backup missing data.

## Fix

Pass the full `backup_paths` slice into `Archiver::archive` and strip each entry against the backup root it actually belongs to (falling back to the entry path unchanged if no root matches). When `--as-path` is used with multiple sources, keep each source as its own subtree under the target to avoid collisions between sources that contain identically-named files.

## Tests

Two regression tests added in `crates/core/tests/integration/backup.rs`:
- `test_backup_multi_source_preserves_files_of_all_sources` (with `--as-path`)
- `test_backup_multi_source_preserves_files_without_as_path` (default mode)

Verified:
```
cargo test -p rustic_core            # 168 passed
cargo test --test integration        # 59 passed
cargo clippy -p rustic_core --all-targets   # no warnings
```

## Notes

This changes the signature of `Archiver::archive` (backup_path `&Path` → backup_paths `&[PathBuf]`). It is `pub`, so downstream users calling it directly would need to update — but the normal entry points (`repo.backup`, `repo.archive`) are unchanged.

### DIFF
--- a/crates/core/CHANGELOG.md
+++ b/crates/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+
+- backup with multiple source paths preserves the files of all sources instead of silently dropping the second source's files (or panicking with `StripPrefixError`) ([rustic#1905](https://github.com/rustic-rs/rustic/issues/1905))
+
 ## [0.13.0](https://github.com/rustic-rs/rustic_core/compare/rustic_core-v0.12.0...rustic_core-v0.13.0) - 2026-08-16
 
 ### Added
--- a/crates/core/src/archiver.rs
+++ b/crates/core/src/archiver.rs
@@ -3,7 +3,7 @@ pub(crate) mod parent;
 pub(crate) mod tree;
 pub(crate) mod tree_archiver;
 
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::thread::scope;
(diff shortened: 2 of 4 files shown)


Lauri-Nomme:
Retracting per maintainer feedback: as-path+multi-source is already rejected in the rustic CLI, and the no-as-path path needs a cleaner verification. Will pursue root-cause analysis and reopen/replace with a validated fix.

PR #561, opened 2026-09-02
feat: count unreadable source files during backup

BradKollmyer:
## Summary

- count source scan and archival errors instead of only logging them
- report those errors through `Progress::error` (`scan` / `archival`)
- attach the source path on local walk/open failures
- expose a `SnapshotSummary.error_count` for the just-completed backup

## Why

restic returns exit code 3 when some source files cannot be read, and still creates a snapshot of the remaining files. rustic currently ignores those errors and exits 0, so automation cannot detect incomplete backups.

## Validation

- `cargo test --test integration test_backup_unreadable_file_sets_error_count` (unix; chmod 000 file)
- `cargo fmt --all`
- clippy on the changed files

A companion rustic CLI PR will use `error_count` to exit 3 and emit restic-compatible JSON `error` / `exit_error` messages.


BradKollmyer:
Companion CLI PR: https://github.com/rustic-rs/rustic/pull/1908

crates/core/src/archiver.rs
```diff
@@ -18,14 +19,21 @@ use crate::{
...
         ReadGlobalIndex,
         indexer::{Indexer, SharedIndexer},
     },
     repofile::{configfile::ConfigFile, snapshotfile::SnapshotFile},
 };
 
+/// Report a source-file error, count it, and continue the backup.
+fn report_source_error(p: &Progress, errors: &AtomicU64, during: &'static str, err: &RusticError) {
+    _ = errors.fetch_add(1, Ordering::Relaxed);
+    let item = err.context_value("path").unwrap_or("");
```
aawsome: instead of the unwrap and later checking for an empty string, please work with an `Option`

crates/core/src/progress.rs
```diff
@@ -87,6 +98,24 @@ pub trait RusticProgress: Send + Sync + 'static + std::fmt::Debug {
...
+    /// The default implementation logs a warning. JSON progress implementations
+    /// emit a restic-compatible error message instead.
+    ///
+    /// # Arguments
+    ///
+    /// * `item` - Usually the path of the problematic file or directory
+    /// * `during` - What was being attempted, e.g. `"scan"` or `"archival"`
+    /// * `message` - Error message
+    fn error(&self, item: &str, during: &str, message: &str) {
+        if item.is_empty() {
```
aawsome: see above.

crates/core/tests/integration/backup.rs
```diff
@@ -307,3 +307,51 @@ fn test_backup_excludes_xattr_entries(set_up_repo: Result<RepoOpen>) -> Result<(
 
     Ok(())
 }
+
+#[cfg(unix)]
+struct RestorePerms<'a>(&'a Path);
+
+#[cfg(unix)]
```
aawsome: Why do we need to restore permissions for a file which is anyway located just in a temp dir?

crates/core/src/repofile/snapshotfile.rs
```diff
@@ -247,6 +247,13 @@ pub struct SnapshotSummary {
 
     /// Total duration that the rustic command ran in seconds
     pub total_duration: f64,
+
+    /// Number of source files/directories that could not be read during this backup run.
+    ///
+    /// This is not persisted in the snapshot file; it is only set on a snapshot
+    /// returned from a just-completed backup.
```
aawsome: Actually I suggest to also persist this information in case the error count > 0. This is then a quite important information...

crates/core/src/archiver.rs
BradKollmyer: Done — `Progress::error` now takes `Option<&str>`, and `report_source_error` passes `err.context_value("path")` through without converting a missing path to an empty string.

crates/core/src/progress.rs
BradKollmyer: Same change here: the trait method now takes `Option<&str>` and the default logger matches on `Some`/`None` instead of checking for an empty string.

crates/core/tests/integration/backup.rs
BradKollmyer: Agreed, it was unnecessary. `unlink` only needs write permission on the parent directory, so the `RestorePerms` drop guard is gone.

crates/core/src/repofile/snapshotfile.rs
BradKollmyer: Now serialized when non-zero (`skip_serializing_if` zero, `default` on deserialize) so existing snapshots stay unchanged and incomplete backups keep the count. The integration test reloads the snapshot from the repo to check it round-trips.

aawsome (approved):
LGTM! Thanks a lot @BradKollmyer

PR #562, opened 2026-09-04
feat: log OpenDAL retries as a single line

BradKollmyer:
## Summary

- replace OpenDAL's default retry interceptor with a compact one-line warning
- keep the same information (error, delay, operation, attempt) without dumping the `Debug` context tree
- match the REST backend style: `Error {err} at {duration:?}, retrying ...`

## Why

Temporary backend errors such as B2 `connection closed before message completed` are already retried. The default interceptor formats the error with `{:?}`, which prints a multi-line Context/Source dump and overwrites progress output. OpenDAL's `Display` form is already a single line.

## Validation

- `cargo test -p rustic_backend --lib opendal`
- `cargo clippy -p rustic_backend --lib -- -D warnings`

A companion rustic CLI PR captures console logs while the TUI owns the terminal so remaining log output cannot overwrite the progress popup.

### DIFF
--- a/crates/backend/CHANGELOG.md
+++ b/crates/backend/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Fixed
+
+- log OpenDAL retries as a single line instead of dumping the error `Debug` tree
+
 ## [0.7.0](https://github.com/rustic-rs/rustic_core/compare/rustic_backend-v0.6.2...rustic_backend-v0.7.0) - 2026-08-16
 
 ### Added
--- a/crates/backend/CHANGELOG.md
+++ b/crates/backend/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 All notable changes to this project will be documented in this file.
 
-## Unreleased
-
-### Fixed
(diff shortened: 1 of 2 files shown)


crates/backend/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Fixed
+
+- log OpenDAL retries as a single line instead of dumping the error `Debug` tree
+
```
aawsome: Can you please remove this? The CHANGELOG gets automatically updated by release_plz when we do a release

aawsome (changes requested):
Hi @BradKollmyer ! Thanks for the PR! LGTM, but please remove the changelog edit - we document changes in the git history and the changelog gets generated.

crates/backend/CHANGELOG.md
BradKollmyer: Removed in b49d4aa. Thanks for the pointer — I will leave CHANGELOG to release-plz from now on.

aawsome (approved):
LGTM! Thanks @BradKollmyer

